PR #27, opened 2019-06-03
feat(field/time-input): expose field and time-input for business applications

efux:
**This is currently WIP. I am not so happy yet with this solution and still looking for a better solution to re-use and to extend the public components for business. I am looking forward to discuss this solution and other (possibly better) solutions with Lukas Spirig on this thursday.**

This PR will export the [Field](https://sbb-angular.app.sbb.ch/develop/content/field) and [Time Input](https://sbb-angular.app.sbb.ch/develop/content/time-input) component as a `@sbb-esta/angular-business` package with the correct styling according to the [SBB Design Guidelines for System Business Applications](https://digital.sbb.ch/en/webapps).

Originally we discussed to export the business components by just re-exporting the modules from the public API with relative paths, to keep the extension as simple as possible. However, ng-packagr does not like relative input paths to directories outside of the root source folder for obvious reasons. Maybe there is a configuration toggle, which I did not try. I tried adding `rootDirs` config, but unfortunately the problem persists.

For making it work, I moved all *.ts and style files to a new folder `projects/angular-components` and left only the package configuration aswell as the specifc styling variables, mixins, ... for business and public in `projects/sbb-esta/angular-business` and `projects/sbb-esta/angular-public`. `angular-components` is the root source folder for both component libraries.

The packages `angular-business` or `angular-public` only contain the exported modules from `public-api.ts` or `business-api.ts`.

The main styles are in `angular-components` as well, however the different styling configuration parameters for Business and Public are stored in their corresponding directories and replaced during the build by include paths.


danielwiehl:
Have you tried using a secondary entry point? It's quite simple and allows you to export some parts of your library under a different sub-namespace, e.g. to avoid name clashes or to better group symbols according to their meaning/usage. 

For instance: 
@sbb-esta/sacl (as the primary entry point)
@sbb-esta/sacl/enterprise (as the secondary entry point)

This mechanism is also used by Angular, e.g. to export `@angular/cdk` as the primary entry point and `@angular/cdk/a11y` as a secondary entry point.

For more information, see https://github.com/ng-packagr/ng-packagr/blob/master/docs/secondary-entrypoints.md.

We successfully use this mechanism in our project-specific widget library to separate its widgets from utilities and RxJS operators.

kyubisation:
@danielwiehl The business/enterprise variants will be contained in a different package. @efux and I previously discussed this offline. We will meet again on thursday to review this PR.

danielwiehl:
Perfect!

efux:
We decided on another approach, I will open a new PR for the components of this PR.

PR #38, opened 2019-06-17
fix(datepicker): prevent auto open of datepicker slave

### LINKED ISSUE #36: Datepicker slave opens when master value changes
When using Datepicker in master/slave setup one currently cannot set the value of the master without having the slave opening the popup. This is a problem when setting the value programmatically.

daniel-sc:
The datepicker slave automatically opens its toggle on multiple triggers.
For some circumstances this is annoying for the user. This change allows
to configure the slave via an input to prevent this behaviour.

closes #36


kyubisation:
This issue should be addressed via #45 
The trigger of the slave is not configureable, but it should be less aggressive. It will now only happen on selecting a date via calendar in the master and when no slave date is set or the slave date is before the newly selected date.

PR #39, opened 2019-06-18
feat(business/button): expose button component for business

efux:
Exposes the `button` component for business applications.

Styling should be according to the [digital guidelines of SBB for business applications](https://digital.sbb.ch/en/webapps/components/button).

The `button` component of the public package has the following modes:
- `primary`
- `secondary`
- `ghost`
- `frameless`

<kbd>public-buttons</kbd>

The `button` component from the business package allows other modes:
- `primary`
- `alternative`
- `ghost`
- `icon` Which should only contain one icon with the `*sbbButtonIcon` directive.

<kbd>business-buttons</kbd>

Business buttons also support icons along with text with the normal modes `primary`, `alternative`, `ghost`, even though this use case was not specifically mentioned in the style guide. However the animations on hover were removed. On hover the background color of the buttons changes the same as without icon.


projects/sbb-esta/angular-business/src/lib/button/button/button.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  Input,
+  ViewEncapsulation
+} from '@angular/core';
+import { BaseButton } from '../../../../../angular-public/src/lib/button/button/base-button';
+
+@Component({
+  // tslint:disable-next-line:component-selector
+  selector: 'button[sbbButton], input[type=submit][sbbButton]',
+  templateUrl: '../../../../../angular-public/src/lib/button/button/button.component.html',
+  styleUrls: ['../../../../../angular-public/src/lib/button/button/button.component.scss'],
```
kyubisation: I'm wondering, if it would make more sense to separate styling a bit more, since there are more differences between public and business styles for buttons. (Icon animation, disabled styling, different button types, ...)
What is your opinion?

efux: I split up the button types to seperate files. For now, I left the arrowbuttons and animations as is. I have to check if this is compliant to the style guide next week.

kyubisation:
@efux LGTM. Thanks for your work. 😃

PR #41, opened 2019-06-19
feat(business/select): expose select component for business

efux:
Exposes the select component for business applications.

Styling should be according to the [digital guidelines of SBB for business applications](https://digital.sbb.ch/en/webapps/components/select), except for the on-hover bar at the left side of the item.

This PR/Branch is based on #32 and therefore the other PR should be merged first.

business-select


efux:
Yes, I think there should be some kind of documentation for the business components, which shows what the components should be doing and how they should look.

I changed the color of the selected item and added the bar on the left side. I originally left it aside (as mentioned in the PR), because it got removed from the public style guide as well. However, I think this might be something which might be added back to the public SACL, because it helps people which cannot see the contrast between red and black well enough to differentiate between them.

business-select2

kyubisation:
@efux LGTM. Thank you very much. As I mentioned I'll try to provide the groundwork for the business section in the showcase. I'll approach you, once I have something ready.

PR #47, opened 2019-06-21
feat(business/datepicker): expose datepicker component for business 

efux:
Exposes the datepicker component for business applications.

Styling should be according to the [digital guidelines of SBB for business applications](https://digital.sbb.ch/en/webapps/components/datepicker).

business-datepicker


kyubisation:
@efux LGTM. Thanks again for your work. 😃

PR #51, opened 2019-06-24
feat(business/checkbox): expose checkbox component for business

efux:
Exposes the checkbox component for business applications.

The business library also offers a styling for indeterminated state.

Styling should be according to the [digital guidelines of SBB for business applications](https://digital.sbb.ch/en/webapps/components/checkbox).


projects/sbb-esta/angular-public/src/lib/checkbox/checkbox/checkbox.component.html
```diff
@@ -16,6 +16,7 @@
 
   <div class="sbb-checkbox-container">
     <sbb-icon-tick class="sbb-checkbox-checked"></sbb-icon-tick>
+    <div *ngIf="!checked" class="sbb-checkbox-indeterminate"></div>
```
kyubisation: This I would like to avoid.
Indeterminate does not exist in public.
Is this maybe possible with pure css? (:content)
However, how can the consumer activate the indeterminate state?
I think extending the component might be the better approach here.

efux: I created a distinct checkbox component for business, which has an indeterminate state.

Please note, that I don't update the indeterminate state of the native element, however, I update ariaChecked. Do you think this is sufficient or should I update the native element as well?

kyubisation: Please create a new template for the business variant, which forwards the indeterminate state to the native checkbox element (<input ... [indeterminate]="indeterminate" ...>).
indeterminate can probably be reduced to a simple property, instead of getter/setter.
You can also attach `@HostBinding('class.sbb-checkbox-indeterminate')` to the indeterminate property. `indeterminateClass` is unnecessary.

PR #53, opened 2019-06-24
fix(business/button): change font opacity for disabled buttons

efux:
This PR adds an opacity to the fonts on disabled buttons for all button types and changes the text color and text color on hover for the ghost button type.


kyubisation:
@efux LGTM. Thanks. 😃

PR #55, opened 2019-06-24
fix(business/field): change gray color to fit style guide

efux:
change the gray color to fit style guide for input field labels and borders from #666 (sbbColorGrey) to #686868 (sbbColorGranite).


kyubisation:
@efux LGTM. Thanks. 😃

PR #59, opened 2019-06-25
fix(business/option): change color of active/hover to red125

efux:
This changes the color of the active option to red, as opposed to as discussed in #41. They adapted the style guide to this, because when using keyboard navigation to select an item the item should be red instead of black.

This affects the select and autocomplete component.


kyubisation:
@efux Was this changed recently? Is that change documented somewhere?

PR #60, opened 2019-06-26
fix(select): change arrow trigger position to absolute

efux:
This PR changes the placement of the arrow trigger in the select component from relative to absolute. This is similar to the arrows for the datepicker component. The reason for this change is, that otherwise the icon is higher than a normal field in the business library and therefore form elements are of different height when placed next to each other.

This change affects the public and business library.

### DIFF
--- a/projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
+++ b/projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
@@ -15,7 +15,7 @@
       }
 
       @include mq($from: desktop5k) {
-        padding-right: 92px;
+        padding-right: toPx($selectTriggerPaddingRight * $scalingFactor5k);
       }
     }
 
@@ -56,7 +56,7 @@
         }
 
         @include mq($from: desktop5k) {
-          right: 28px;
+          right: toPx($selectArrowRight * $scalingFactor5k);
           width: toPx($selectArrowWidthHeight * $scalingFactor5k);
           height: toPx($selectArrowWidthHeight * $scalingFactor5k);
         }
--- a/projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
+++ b/projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
@@ -43,11 +43,12 @@
       @include absoluteCenterY();
       right: toPx($selectArrowRight);
 
-      @include svgIconColor($sbbColorGrey);
       width: toPx($selectArrowWidthHeight);
       height: toPx($selectArrowWidthHeight);
 
       @include publicOnly() {
+        @include svgIconColor($sbbColorGrey);
+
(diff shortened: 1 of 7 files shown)


projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
```diff
@@ -3,10 +3,21 @@
...
-    padding-right: $selectTriggerPaddingRight;
+    padding-right: toPx($selectTriggerPaddingRight);
+
+    @include publicOnly() {
+      @include mq($from: desktop4k) {
+        padding-right: toPx($selectTriggerPaddingRight * $scalingFactor4k);
+      }
+
+      @include mq($from: desktop5k) {
+        padding-right: 92px;
```
kyubisation: Usually the desktop5k variant is simply calculated with the $scalingFactor5k. Why is this not the case here?

projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
```diff
@@ -29,23 +40,45 @@
...
+        @include svgIconColor($sbbColorGrey);
+
         @include mq($from: desktop4k) {
+          right: toPx($selectArrowRight * $scalingFactor4k);
           width: toPx($selectArrowWidthHeight * $scalingFactor4k);
           height: toPx($selectArrowWidthHeight * $scalingFactor4k);
         }
 
         @include mq($from: desktop5k) {
+          right: 28px;
```
kyubisation: Same question as for padding-right.

projects/sbb-esta/angular-public/src/lib/select/styles/_select-base.scss
```diff
@@ -3,10 +3,21 @@
 @mixin selectBase {
   .sbb-select-trigger {
     @include inputfields;
+    position: relative;
     cursor: pointer;
     display: flex;
     align-items: center;
-    padding-right: $selectTriggerPaddingRight;
+    padding-right: toPx($selectTriggerPaddingRight);
```
kyubisation: padding-right should also be applied for native select.

projects/sbb-esta/angular-public/src/styles/typography/_form.scss
```diff
@@ -78,11 +80,13 @@
   @include publicOnly() {
     background: url($selectNativeBackground) 97.5% no-repeat;
```
kyubisation: The padding-right change displaced the position of the arrow.
This can be solved by removing the percentage from background (and moving it ouside publicOnly/businessOnly) and adding `background-position: center right $selectArrowRight` (also replace `background-position-x` from the 4k/5k with `background-position: toPx($selectArrowRight * $scalingFactor4k/5k)`)

kyubisation:
LGTM. Thanks again. 😃

PR #66, opened 2019-07-01
feat(business): add possibility for themes

efux:
This PR is a proposal for adding the possibility to change the SBB red to other colors for theming purposes.

The overrides are done by selecting the elements more specific than the default styling. Because of this the overrides are very verbose. This expects the classes and tag names to remain stable throughout the development of the libraries. However, for the elements currently present in the business library, the overrides in SCSS files from this PR should already be sufficient.

The overrides are packaged together with the business library, but can be included only when needed. The developer, who wants to change the default color, would include something like the following in his main style sheet:

```scss
@import '../../../../node_modules/@sbb-esta/angular-business/typography.css';

$themePrimaryColor: yellow;
$themePrimaryColorHover: orange;

@import '../../../../node_modules/@sbb-esta/angular-business/overrides';
```


timschoch:
I saw this pull request when browsing the repo and we're looking forward to some form of theming. Have you already decided what parts will be themable? (eg. Gutters, Typography...)

For our cargo client portal we want to stick mostly to default styling for layout stuff since it should be inline with the rest of the public appearance. But for the data heavy stuff like tables we'll use the business styling. To fine tune these two worlds I can see it would be helpful to have more control over the design.

Also, having access to more SCSS Variables would help with styling our own components. Right now, we need to duplicate those variables to our own scss file, whichs is a really bad but necessary thing.

kyubisation:
@timschoch Could you open a [feature request](https://github.com/SchweizerischeBundesbahnen/sbb-angular/issues/new?assignees=&labels=enhancement&template=feature_request.md&title=Feature+request%3A+) with details of what you need and what you would like?
I'm still working on the architecture of theming and I am not yet convinced, that this PR is the right approach. I would like the discussion (and specification) to be in a dedicated issue.

efux:
Discussion is in #70.

PR #67, opened 2019-07-01
fix(checkbox): fix change behavior for checkboxes

kyubisation:


projects/sbb-esta/angular-business/src/lib/checkbox/checkbox/checkbox.component.html
```diff
@@ -1,18 +1,22 @@
...
     [id]="inputId"
-    [value]="value"
+    [attr.name]="name"
+    [attr.value]="value"
     [required]="required"
     [checked]="checked"
     [indeterminate]="indeterminate"
-    [attr.aria-checked]="ariaChecked"
+    [tabIndex]="tabIndex"
+    [attr.aria-checked]="checked"
```
efux: Should probably be

```html
[attr.aria-checked]="ariaChecked"
```

or something else, if the checkbox has state indeterminate it should be mixed.

kyubisation: Thanks. I missed that.

PR #99, opened 2019-07-12
feat(business/accordion): expose accordion for business

### LINKED ISSUE #77: feat(business): accordion styling
Design Spec: https://digital.sbb.ch/en/webapps/components/accordion

Reuse the public accordion and apply business styling.
See business variant of autocomplete for reference.

efux:
Exposes the accordion component for business applications. Styling should be accoring to the [SBB Style Guide for Business Applications](https://digital.sbb.ch/en/webapps/components/accordion).

For the toggle I used a background-svg similar to the native select, for not needing to tamper with the html code of the components. Otherwise I would have to extend multiple components just to exchange the icon.

Closes #77

business-accordion

### DIFF
--- a/projects/sbb-esta/angular-business/src/lib/accordion/accordion.ts
+++ b/projects/sbb-esta/angular-business/src/lib/accordion/accordion.ts
@@ -1,0 +1,1 @@
+export * from '../../../../angular-public/src/lib/accordion/accordion';
--- a/projects/sbb-esta/angular-business/src/public-api.ts
+++ b/projects/sbb-esta/angular-business/src/public-api.ts
@@ -1,6 +1,7 @@
 /*
  * Public API Surface of angular-business
  */
+export * from './lib/accordion/accordion';
 export * from './lib/autocomplete/autocomplete';
 export * from './lib/button/button';
 export * from './lib/checkbox/checkbox';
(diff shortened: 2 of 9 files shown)


efux:
Currently does not work in IE yet. Please wait with merging.

efux:
I fixed the IE bug. For doing this I had to encode the inline svg for the background. I think, this is a problem that is present for other elements in public too. E.g. [native select](https://sbb-angular.app.sbb.ch/latest/content/select) is affected in IE.

I am not to happy with the approach with the background svg I chose here, because it also contains the plus and minus icons in the template, which are not needed. What do you think? Good enough or should I extend all components of the accordion to replace the templace from expansion-panel-header.

kyubisation:
Native select will be tracked in #102. If you have an idea on how to easily fix this, I would appreciate a PR.
Considering the css svg approach; An option would be to also move the plus and minus icons to css. What are your thoughts on the matter?

efux:
I think moving the other two icons to CSS is not optimal, but a step in the right direction. Especially since they are purely visual features and do not offer any functionality on their own. I will update my PR.

efux:
I replaced the html template, so that all the indicators for the accordion are css-only (business and public).

kyubisation:
LGTM. Thanks for you work. 😃

PR #109, opened 2019-07-16
fix(business/checkbox): add sbb-checkbox class to public and business

efux:
Moves the binding of the class sbb-checkbox to the base checkbox component. Otherwise the correct styling is not applied to business components.

If possible, can you please create a patch-release of the business package after merging.


kyubisation:
Thanks! LGTM.
We urgently need the business section in the showcase. I'll see that I can reserve some time on friday to set it up.

Issue #124, opened 2019-08-07
Bug: sbb-usermenu shifts username on clicking


jeripeierSBB:
**Describe the bug**
if a user is logged in an clicks on his username (sbb-usermenu) (to expand the dropdown), the avatar/initials and the name is shifted.

**To Reproduce**
1. Go to https://sbb-angular.app.sbb.ch/latest/content/usermenu
2. Click on Login
3. Click on username
4. username and avatar are shifted instead of staying at its place

**Expected behavior**
avatar and username stay at its place

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome

**Additional context**
We "solved" the problem by appending following styles inside our application:

```
:host {
  ::ng-deep {
    sbb-usermenu {
      .sbb-usermenu-logged-in {

        /*
          if username is shown and dropdown collapsed -> correct margins of box
         */
        &.sbb-dropdown-trigger:not(.sbb-usermenu-logged-in-active) .sbb-usermenu-logged-in-collapsed {
          border-left: 2px solid transparent;
          margin-top: 2px;
        }

        /*
          if username is shown and dropdown collapsed -> correct margin of user name
        */
        .sbb-usermenu-logged-in-collapsed:not(.sbb-usermenu-logged-in-collapsed-dropdown-expanded) .sbb-usermenu-logged-in-expanded-info-user .sbb-usermenu-logged-in-expanded-displayName {
          margin-top: -2px;
        }

        /*
          if username is shown and dropdown is expanded -> shift up user name and icon
         */
        .sbb-usermenu-logged-in-collapsed-dropdown-expanded {
          margin-top: 0;
        }
      }
    }
  }
}
```

kyubisation:
Closed by #128

jeripeierSBB:
@kyubisation For me this bug is still present, see video: ![ezgif-4-182d05c07191](https://user-images.githubusercontent.com/29278232/64966606-648ef180-d89f-11e9-8870-64af78da046f.gif)

PR #128, opened 2019-08-13
fix(usermenu): fix layout after click (#124)

daniele92vp:


projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
```diff
@@ -15,6 +15,11 @@ $borderActive5K: solid toPx(4 * $scalingFactor5k) $sbbColorGrey;
   width: 288px;
   height: 48px;
   line-height: 1.533;
+  border-top: 2px solid transparent;
```
kyubisation: This is equivalent to `border: 2px solid transparent;`

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
```diff
@@ -15,6 +15,11 @@ $borderActive5K: solid toPx(4 * $scalingFactor5k) $sbbColorGrey;
...
+  border-left: 2px solid transparent;
+  border-right: 2px solid transparent;
+  border-bottom: 2px solid transparent;
+  outline: 0;
```
kyubisation: Unless there is an alternative hightlight on focus, hiding outline is a problem for accessibility. So this should be removed.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
```diff
@@ -31,17 +36,20 @@ $borderActive5K: solid toPx(4 * $scalingFactor5k) $sbbColorGrey;
   border-top: $borderActive;
   border-right: $borderActive;
   border-left: $borderActive;
+  border-bottom: $borderActive;
```
kyubisation: What is the reason for adding this? I see no difference in testing?
(Also replace with `border: $borderActive;` if this is required?)
This also applies to the border-bottom lines below.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
```diff
@@ -50,7 +58,6 @@ $borderActive5K: solid toPx(4 * $scalingFactor5k) $sbbColorGrey;
   align-items: center;
   margin-left: 3px;
   margin-top: 3px;
-  padding-left: 16px;
```
kyubisation: Removing this will cause the icon to be to close to the border when the dropdown is open.
It might make sense to move this to the `usermenuLoggedInActive` mixin.
Be aware that this is potentially a ui breaking change for consumers. Be very carefull with these types of changes.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
daniele92vp: Yes you are totally right I will change to border: 2px solid transparent. 
And will also remove outline: 0.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
daniele92vp: If you do not put this, When the curtain goes up the bottom edge is missing.

usermenu-up-issue

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
daniele92vp: This actually solve the issue that was highlighted, when the user click on login everything moves to the right because of this padding-left: 16px. So let me know if is an expected behavior or not.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
kyubisation: Good catch. Please reduce to `border: $borderActive;`. Also apply for the changes below.

projects/sbb-esta/angular-public/src/lib/usermenu/_usermenu.scss
kyubisation: The specification (https://digital.sbb.ch/de/websites/components/usermenu) specifies no padding  on collapsed state and padding on dropdown state.
Simply removing the padding will fix the issue in collapsed state but cause a problem in the dropdown state.

Issue #120, opened 2019-07-31
Bug: sbb-toggle state can't be set from outside


jeripeierSBB:
**Describe the bug**
It's not possible to control <sbb-toggle> from outside. So if there is a default value in a form, it's not reflected by the component.

**To Reproduce**
1. Go to https://stackblitz.com/edit/angular-sbb-rbevqk
2. "Resolved" should be selected, but it's not
3. Read logs for more information

**Expected behavior**
correct toggle-option is set as active and is shown highlighted to the user

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome

PR #130, opened 2019-08-13
fix(toggle): accept initial value for state (#120)

gianluca-r:


projects/sbb-esta/angular-public/src/lib/toggle/toggle/toggle.component.ts
```diff
@@ -112,7 +112,13 @@ export class ToggleComponent extends RadioButton
     this._zone.onStable.pipe(first()).subscribe(() =>
       this._zone.run(() => {
         this._checkNumOfOptions();
-        const defaultOption = this.toggleOptions.toArray()[0];
+        // Before assigning the first tab to defaultOption, I check whether a different value has been specified
+        let defaultOption = this.toggleOptions
```
kyubisation: This could optionally be shortened to 
```ts
const defaultOption = this.toggleOptions
  .toArray()
  .find(toggle => toggle.value === this.value) || this.toggleOptions.toArray()[0];
```

kyubisation:
Thank you for the PR. 😃 
I have suggested some minor changes.

kyubisation:
LGTM
Thank you very much 😃

Issue #122, opened 2019-08-01
Bug: sbb-table tableAlignment not working for header in IE11


daniel-sc:
**Describe the bug**

**To Reproduce**
Go with IE11 to https://angular.app.sbb.ch/latest/content/table -> Examples and select "left" for "Text Alignment".

**Expected behavior**
All cells should be aligned left, as e.g. in Chrome:
![image](https://user-images.githubusercontent.com/117919/62303477-7a299280-b47c-11e9-89f8-235770b439f5.png)

**Screenshots**
In IE it looks as follows (red markup indicating the problem):
![image](https://user-images.githubusercontent.com/117919/62303534-99282480-b47c-11e9-8699-a72211e4b199.png)

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser IE
 - Version 11

PR #133, opened 2019-08-16
fix(table): tableAlignment works on table headers in IE11 (#122)

gianluca-r:


kyubisation:
LGTM 😃 
Thank you very much

Issue #125, opened 2019-08-08
Bug: sbb-toggle styling problem in IE11


jeripeierSBB:
**Describe the bug**
content on sbb-toggle is always displayed in IE11, event when there is no content (subsection below toggle labels) (See screenshot).

**To Reproduce**
1. Go to 'https://sbb-angular.app.sbb.ch/latest/content/toggle.' in IE11
2. See that toggle content is always displayed

**Expected behavior**
Same behavior as in other browsers

**Screenshots**
![2019-08-08 08_09_50-SBB Angular - Internet Explorer](https://user-images.githubusercontent.com/29278232/62679179-f7d92b00-b9b3-11e9-829c-6cdf4e9bb184.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser IE
 - Version 11

PR #134, opened 2019-08-16
fix(toggle): use for-loop in place of forEach on NodeList (#125)

gianluca-r:
IE11 doesn't support .forEach on NodeLists, which was causing the issue.
An alternative could be adding a generic polyfill for the issue.


kyubisation:
LGTM 😃 
Thank you very much

Issue #115, opened 2019-07-24
Bug: sbb-file-selector model changes not displayed


jeripeierSBB:
**Describe the bug**
If the model of <sbb-file-selector> changes, the view of the file list is not being updated immediately.

**To Reproduce**
1. Go to https://stackblitz.com/edit/angular-sbb-owhuru
2. Upload a file
3. Filelist should be reset by controller, but the file list in the view is not changed
4. Click somewhere in the filelist and the view is updated (file not displayed anymore)

**Expected behavior**
Changes to model should be displayed immediately

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser chrome

PR #135, opened 2019-08-16
fix(file-selector): fix upload with two way binding update (#115)

daniele92vp:
The test has been done on the changeDetector event. This will make sure that angular will detect any changes made also during the same cycle.


kyubisation:
LGTM 😃 
Thank you very much

Issue #136, opened 2019-08-16
Bug: toggleChange on sbb-toggle is fired 4 times per toggle


gianluca-r:
**Describe the bug**
When toggling buttons on the toggle component, 4 toggleChange events are fired, which definitely seems to be unintended behaviour.

**To Reproduce**
1. Go to https://stackblitz.com/edit/angular-sbb-vjvqfv
2. Open the console
3. Click on 'resolved' toggle button
4. Console prints 4 events, alternating between button 1 & 2

**Expected behavior**
The toggleChange event on sbb-toggle should fire once when the component is toggled.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome/Firefox

gianluca-r:
To add to this, the current toggleChange event is fired once per toggle button, which means it is fired twice every time, once per toggle button value.

![image](https://user-images.githubusercontent.com/33316356/63258728-feea1e00-c27c-11e9-9bce-be5f53f011bd.png)

This event by itself doesn't give any details on whether the tab was actually selected or not. I suggest changing the event data to something more meaningful than simply the toggle option value, and use instead an event interface.

```
{
    value: any;
    isSelected: boolean;

    ....
}
```

PR #137, opened 2019-08-19
fix(toggle): fire toggleChange only when status actually changes (#136)

gianluca-r:


kyubisation:
LGTM 😃 
Thank you very much

PR #139, opened 2019-08-20
fix(business/accordion): remove media query breakpoints

efux:
This fixes an unwanted jump in padding when the screen gets small for the business theme.


kyubisation:
LGTM 😃 
Thank you very much

Issue #141, opened 2019-08-22
Bug: Radio Button circle shrinks on small screens


bollerdominik:
**Describe the bug**
Radio Button circle shrinks on small screens

**To Reproduce**
Use a sbb-radio-button in a small screen or within a small container.
See:
https://stackblitz.com/edit/angular-sbb-l954ew

**Expected behavior**
Radio button circle keeps the same size on all screen sizes

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox, Edge, IE, Chrome

efux:
This also happens with checkboxes. See https://stackblitz.com/edit/angular-sbb-rzg7ws

PR #142, opened 2019-08-22
fix(business/checkbox): fill block of indeterminate

efux:
This shows the indeterminate state of the business checkbox as block with background color. If someone tries to resize the checkbox, this is necessary to prevent a wrong look. This is precautionary.

### DIFF
--- a/projects/sbb-esta/angular-public/src/lib/checkbox/_checkbox.scss
+++ b/projects/sbb-esta/angular-public/src/lib/checkbox/_checkbox.scss
@@ -65,6 +65,7 @@ $checkBoxContainerBorder: 2;
           left: 4px;
           top: 8.5px;
           border: 1px solid $sbbColorGranite;
+          background-color: $sbbColorGranite;
           display: block;
         }
 


projects/sbb-esta/angular-public/src/lib/checkbox/_checkbox.scss
```diff
@@ -65,6 +65,7 @@ $checkBoxContainerBorder: 2;
           left: 4px;
           top: 8.5px;
           border: 1px solid $sbbColorGranite;
+          background-color: $sbbColorGranite;
```
kyubisation: Why was this necessary?

efux: This might be related to #141 . On my screen the checkbox was bigger then it should be and the indeterminate state was displayed only as a grey border instead of a filled grey box. This might be  redundant, when #141 is fixed and the checkbox always has the pre-defined size.

efux:
I thought this might be related to #141, however #231 did not fix it so I investigated further and found the issue. It was not related to this library at all, instead somewhere `box-sizing: border-box` was overwritten.

Therefore I close this PR.

Issue #93, opened 2019-07-05
feat(business): header


kyubisation:
Design Spec: 
- https://digital.sbb.ch/en/webapps/modules/header
- https://digital.sbb.ch/en/webapps/modules/mainnavigation

API Spec:

- Selector: sbb-header
- Inputs:
  - label: string
  - subtitle?: string
  - environment?: string
  - environmentColor?: string
- Should support a tag and button tag in content projection
- Should support sbb-usermenu via content projection
- Should support an optional custom brand logo. If no logo is given, should fall back to SBB brand logo. (Use content projection with "[brand],.brand")

Examples
```html
<sbb-header [label]="label">
  <a routerLink="/">Home</a>
  <a routerLink="/navigation1">Navigation 1</a>
  <button type="button" [sbbDropdown]="dropdown">Navigation 4</button>
  <sbb-dropdown #dropdown="sbbDropdown">
    <a sbbDropdownItem routerLink="/navigation4/section1" routerLinkActive="sbb-selected">Section 1</a>
    <a sbbDropdownItem routerLink="/navigation4/section2" routerLinkActive="sbb-selected">Section 2</a>
  </sbb-dropdown>
</sbb-header>
```
```html
<sbb-header [label]="label">
  <a routerLink="/">Home</a>
  <a routerLink="/navigation1">Navigation 1</a>
  <button type="button" [sbbDropdown]="dropdown">Navigation 4</button>
  <sbb-dropdown #dropdown="sbbDropdown">
    <a sbbDropdownItem routerLink="/navigation4/section1" routerLinkActive="sbb-selected">Section 1</a>
    <a sbbDropdownItem routerLink="/navigation4/section2" routerLinkActive="sbb-selected">Section 2</a>
  </sbb-dropdown>
  <sbb-usermenu [userName]="user.userName" [displayName]="user.displayName" (loginRequest)="login()">
    <img class="image" sbbIcon src="assets/images/user-avatar.png" />
    <sbb-dropdown>
      <button sbbDropdownItem type="button" (click)="logout(user)">Logout</button>
    </sbb-dropdown>
  </sbb-usermenu>
</sbb-header>
```
```html
<sbb-header [label]="label">
  <a routerLink="/">Home</a>
  <a routerLink="/navigation1">Navigation 1</a>
  <button type="button" [sbbDropdown]="dropdown">Navigation 4</button>
  <sbb-dropdown #dropdown="sbbDropdown">
[...]

PR #148, opened 2019-08-27
feat(business/header): implement header component

gianluca-r:
Small changes to dropdown-trigger were made to allow it to automatically resize when the host component changes it's width, if no panelWidth is set, otherwise it wouldn't register and stay the old size inside the header.

Closes (#93).


projects/angular-showcase/src/app/examples/examples.module.ts
```diff
@@ -44,6 +42,8 @@ import {
   UserMenuModule
 } from '@sbb-esta/angular-public';
 
+import { HeaderModule } from '../../../../sbb-esta/angular-business/src/lib/header/header.module';
```
kyubisation: This should be imported from `@sbb-esta/angular-business`

projects/sbb-esta/angular-business/src/lib/header/header/header.component.html
```diff
@@ -0,0 +1,42 @@
...
+<div class="sbb-header-usermenu">
+  <ng-content select="sbb-usermenu"></ng-content>
+</div>
+
+<!-- If an icon is provided, use that, otherwise use the sbb logo -->
+<!-- TODO: make cleaner -->
+<div #iconContent class="sbb-header-logo">
+  <ng-content select="[brand],.brand"></ng-content>
+</div>
+<div *ngIf="!isIconProvided" class="sbb-header-logo">
```
kyubisation: You can replace !isIconProvided with !iconContent.childElementCount  and remove the iconContent property in the .ts file

projects/sbb-esta/angular-business/src/lib/header/header/header.component.scss
```diff
@@ -0,0 +1,84 @@
...
+  width: 100%;
+
+  position: fixed;
+  top: 0px;
+  left: 0px;
+  z-index: 1000;
+  height: 54px;
+
+  background-color: white;
+  border-bottom: solid 1px #dcdcdc;
```
kyubisation: Use the colors from _colors.scss (import via `@import 'common';`)

projects/sbb-esta/angular-business/src/lib/header/header/header.component.scss
```diff
@@ -0,0 +1,84 @@
...
+    position: absolute;
+    top: 12px;
+    left: -20px;
+    text-align: center;
+    line-height: 12px;
+    letter-spacing: 0px;
+    font-size: 10px;
+    color: #f0f0f0;
+    transform: rotate(-45deg);
+    -webkit-transform: rotate(-45deg);
```
kyubisation: Browser specific variants are not needed, since postcss will add them where necessary

projects/sbb-esta/angular-business/src/lib/header/header/header.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class HeaderComponent implements OnInit {
+  /**
+   * Css class of a sbb-header
+   */
+  @HostBinding('class.sbb-header') cssClass = true;
```
kyubisation: Change comment to /** @docs-private */, since this is an internal detail

projects/sbb-esta/angular-business/src/lib/header/styles/_navButton.scss
```diff
@@ -0,0 +1,101 @@
+@mixin navButton() {
```
kyubisation: Merge this into the navbutton.component.scss (Originally we had a different use case for these style files, but that is no longer necessary. The structure is merely legacy.)

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
```diff
@@ -0,0 +1,80 @@
...
+  Component,
+  ElementRef,
+  HostBinding,
+  OnDestroy,
+  ViewChild,
+  ViewEncapsulation
+} from '@angular/core';
+
+@Component({
+  selector: 'sbb-navbutton',
```
kyubisation: I would like this to be refactored, so the consumer must use this.
Change the selector to 'button[sbbNavbutton]' and refactor the usage.
I realize this is a rather large change, but I believe this reduces complexity.
I apologize for this.

Usage by consumer
```
<sbb-header ...>
  ...
  <button sbbNavbutton type="button" [sbbDropdown]="dropdown">Navigation 4</button>
  ...
</sbb-header>
```

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
```diff
@@ -0,0 +1,80 @@
...
+   * Observer to detect when an eventual dropdown attached to childButton is expanded or collapsed.
+   * Used to propagate change detection.
+   */
+  private _attributeObserver: MutationObserver;
+
+  constructor(private _changeDetectorRef: ChangeDetectorRef) {}
+
+  ngAfterViewInit() {
+    this._checkChild();
+    this._attributeObserver = new MutationObserver(() => this._changeDetectorRef.markForCheck());
```
kyubisation: Prefer usage of https://material.angular.io/cdk/observers/api, since this already considers edge cases

projects/sbb-esta/angular-public/src/lib/dropdown/dropdown-trigger.directive.ts
```diff
@@ -585,4 +592,22 @@ export class DropdownTriggerDirective implements OnDestroy {
     const element = this._elementRef.nativeElement;
     return !element.readOnly && !element.disabled && !this.dropdownDisabled;
   }
+
+  private _attachObserverToHost() {
+    // Make sure we don't have more than one
+    this._unsubscribeHostObserver();
+
+    this._hostObserver = new MutationObserver(() => {
```
kyubisation: Prefer usage of https://material.angular.io/cdk/observers/api

projects/sbb-esta/angular-business/src/lib/header/styles/_header.scss
```diff
@@ -0,0 +1,12 @@
+//@import 'common';
+@import '../../../../../angular-public/src/styles/common/common';
```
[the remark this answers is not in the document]
gianluca-r: My bad, I was using relative imports for quicker testing and forgot to revert them back.

projects/angular-showcase/src/app/examples/examples.module.ts
gianluca-r: Same as above, my bad.

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
gianluca-r: Very interesting, didn't know about these, looks like they'll help handling those cases even better.

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
gianluca-r: No problem, I'll see what I can do to adjust it to that spec.

projects/sbb-esta/angular-business/src/lib/header/header/header.component.html
gianluca-r: I had tried that in the beginning and had some issues with undefined values, but probably it was due to side effects which I should've solved by now.
Regardless, I'll remove the property.

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
```diff
@@ -29,52 +33,67 @@ export class NavbuttonComponent implements AfterViewInit, OnDestroy {
...
+
+  /**
+   * Css width to apply when the dropdown is opened.
+   */
+  dropdownWidth = '200px';
+
+  /** @docs-private */
+  private _subscriptions: Subscription[] = [];
+
+  constructor(@SkipSelf() private _changeDetectorRef: ChangeDetectorRef, private _el: ElementRef) {}
```
kyubisation: You should be able to optionally inject DropdownTriggerDirective: @Optional() private _dropdownTrigger: DropdownTriggerDirective instead of the ContentChild

projects/angular-showcase/src/app/examples/header-showcase/header-showcase.component.html
```diff
@@ -4,10 +4,12 @@
   [environment]="'dev'"
   [environmentColor]="'darkgreen'"
 >
-  <a routerLink="/">Home</a>
-  <a routerLink="/navigation1">Another module</a>
-  <button type="button" [sbbDropdown]="dropdown">Navigation 1</button>
-  <button type="button">Navigation 2</button>
+  <a sbbNavbutton routerLink="/">Home</a>
```
kyubisation: Is it necessary to use the navbutton component for a tags? If possible, I would like to only have to use it for buttons.

projects/sbb-esta/angular-business/src/lib/header/header/header.component.ts
```diff
@@ -64,51 +59,38 @@ export class HeaderComponent implements OnInit {
...
   ngOnInit() {
     this._checkLabel();
-    const element = this.ngContent.nativeElement;
-    this._componentFactory = this._resolver.resolveComponentFactory(NavbuttonComponent);
+  }
 
-    while (element.children.length > 0) {
-      this._addElementToNavbar(element.children[0]);
+  ngAfterViewInit() {
+    // Absolute positioning of buttons so that they're all 70px apart and won't
```
kyubisation: I'm still thinking about an alternative for this, but I think that would require a change to the dropdown implementation...
It's fine for now

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
```diff
@@ -29,52 +33,67 @@ export class NavbuttonComponent implements AfterViewInit, OnDestroy {
...
-      this._attributeObserver.disconnect();
+    while (this._subscriptions.length) {
+      this._subscriptions.shift().unsubscribe();
     }
   }
 
   /**
    * Returns whether childButton has a dropdown attached.
    */
   get isDropdown() {
```
kyubisation: Is this getter still required?

projects/sbb-esta/angular-business/src/lib/header/navbutton/navbutton.component.ts
```diff
@@ -29,52 +33,67 @@ export class NavbuttonComponent implements AfterViewInit, OnDestroy {
...
+  @ContentChild(DropdownTriggerDirective, { static: true })
+  dropdownTrigger: DropdownTriggerDirective;
 
-  constructor(private _changeDetectorRef: ChangeDetectorRef) {}
+  /**
+   * Returns whether childButton's dropdown is expanded.
+   */
+  private _isDropdownExpanded = false;
+  @HostBinding('class.sbb-navbutton-dropdown-expanded')
+  get isDropdownExpanded() {
```
kyubisation: If this getter is only necessary for styling, you can use attribute selectors. In this case the current element has the attribute aria-expanded="true" when open.

e.g.
```css
.sbb-navbutton[aria-expanded="true"] { ... }
```

projects/angular-showcase/src/app/examples/header-showcase/header-showcase.component.html
gianluca-r: No, I thought you meant to have it for all of them, but it's a minor change to have it for buttons only.

projects/angular-showcase/src/app/public/examples/examples.module.ts
```diff
@@ -44,6 +42,8 @@ import {
   UserMenuModule
 } from '@sbb-esta/angular-public';
 
+import { HeaderModule } from '@sbb-esta/angular-business';
```
kyubisation: Move this import to the business examples

projects/sbb-esta/angular-business/src/lib/header/header.ts
```diff
@@ -0,0 +1,3 @@
+export * from './header.module';
```
kyubisation: Could you add a simple documentation in a header.md file?

kyubisation:
LGTM
@gianluca-r Thank you very much 😃

Issue #22, opened 2019-05-28
Feature Request: Clear Input fields


kyubisation:
Copied from https://issues-ext.sbb.ch/browse/SACL-141

If Input fields are used as Search/Filters, it would be helpful to have a simple way to clear the filter and reset it to the default value. Right now, the user is forced to reset every field by clicking the field, selecting the content and removing it. (Or selecting the default value in a select field)

 

**Requirements:**

- once a value has been set, offer a consistent way to reset the filter with one-click
- a default value can be provided to which the filter resets.

 

FYI, we currently use these components as filters:
https://sbb-angular.app.sbb.ch/develop/content/autocomplete
https://sbb-angular.app.sbb.ch/develop/content/datepicker
https://sbb-angular.app.sbb.ch/develop/content/field
https://sbb-angular.app.sbb.ch/develop/content/search

kyubisation:
![image](https://user-images.githubusercontent.com/594745/58462840-faf1a880-8132-11e9-96ff-acd02b89166a.png)

timschoch:
Any updates on this? (maybe @frenggi)

kyubisation:
Openend a feature request upstream: sbb-design-systems/sbb-design-system#7
Sorry for the delay in answering.
From a technical standpoint, this is a relatively simple thing to implement, although there are a few things to consider.

```html
<input type="text" formControlName="example" #ref><sbb-clear-input input="ref"></sbb-clear-input>
```
```html
<input type="text" [(ngModel)]="example" #ref><sbb-clear-input input="ref"></sbb-clear-input>
```

sonatdev:
Hi! Do you have any updates on this?

sandrooco:
@kyubisation & @Frenggi Is this currently scheduled? Cheers!

kyubisation:
Hello @sandrooco 
Unfortunately this has very low priority at the moment. It is still part of our backlog, but I do not think that we will be able to tackle this in the near future.

PR #150, opened 2019-08-30
feat(clear-input): implement clear input component

gianluca-r:
Closes (#22).


jeripeierSBB:
Thank you for the PR. Unfortunately we are going to close it, due it's being out of date and not aligned with our architecture and design goals. We are still open for suggestions but this should probably be implemented in the form-field module.

PR #174, opened 2019-09-17
fix(icons): set focusable to false for svg icons

### LINKED ISSUE #155: Bug: DatePicker has incorrect Tab behaviour in IE11
**Describe the bug**
When using 'Tab' to iterate through a DatePicker with a Toggle (the calendar icon) in IE11, the tab disappears once before and once right after focusing on the Toggle. Firefox and Chrome work both as intended

**To Reproduce**
1. Go to 'https://angular.app.sbb.ch/latest/content/datepicker'
2. Click on 'Examples'
3. Tap through the datepickers with the toggle activated 
4. Tab is lost in IE11

**Expected behavior**
Correct tab behaviour without 'nirvana tabs' ;)

 - OS: Windows
 - Browser IE11
 - Version 11.885.17134.0

efux:
This should fix an issue on IE 11, which leads to a weird sequence when tabbing over the datepicker or buttons with icons.

It would be highly appreciated, if we could include such a fix in a release as soon as possible.

Should close #155


projects/sbb-esta/angular-icons/src/lib/arrow/icon-arrows-circle.module.ts
```diff
@@ -13,18 +13,20 @@ import { IconBase } from '../icon-base';
...
         *ngSwitchDefault
+        focusable="false"
         [attr.class]="'sbb-svg-icon ' + svgClass"
         xmlns="http://www.w3.org/2000/svg"
         viewBox="0 0 24 24"
       >
         <path
           fill="none"
           stroke="#000"
-          d="M4.51 12a7.5 7.5 0 0 1 7.514-7.486c2.912.006 6.238 2.346 7.466 4.986m.01-4.49l-.01 4.49H15m4.5 2.5a7.51 7.51 0 0 1-7.539 7.48c-2.894-.01-6.214-2.353-7.451-4.97M4.5 19l.01-4.49L9 14.5"
```
efux: Is this an issue? Why did it do that?

efux: It does this after generating the icons with npm run build:icons and then npm run format.

kyubisation:
It appears there really is not a difference and prettier seems to be able to optimize the svg even more, if only a little bit.
So, LGTM. Thank you for the contribution! 😃

PR #178, opened 2019-09-19
Enhancements custom components for header

daniel-sc:
This resolves the following problems with the header:

1. use a custom component for the user menu (e.g. which wraps the sbb-usermenu)
2. add additional components to the header (e.g. auto save feedback icons)

### DIFF
--- a/projects/sbb-esta/angular-business/header/src/header/header.component.html
+++ b/projects/sbb-esta/angular-business/header/src/header/header.component.html
@@ -21,24 +21,27 @@
   </div>
   <template #navigationButtons></template>
 </div>
-<div class="sbb-header-additional-content">
-  <ng-content select=".additional-header-content"></ng-content>
-</div>
-<div class="sbb-header-usermenu">
-  <ng-content select="sbb-usermenu,.sbb-usermenu"></ng-content>
-</div>
 
-<!-- If an icon is provided, use that, otherwise use the sbb logo -->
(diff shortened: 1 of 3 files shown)


kyubisation:
@daniel-sc In the future, could you please follow the commit message guidelines also for the naming of the PRs? This would make it easier for me to merge.
Nevertheless thank you for the contribution. I will now have a look at it. 😃

projects/sbb-esta/angular-business/header/src/header/header.component.html
```diff
@@ -21,21 +21,27 @@
...
+  <div class="sbb-header-additional-content">
+    <ng-content select=".additional-header-content"></ng-content>
+  </div>
+  <div class="sbb-header-usermenu">
+    <ng-content select="sbb-usermenu,.sbb-usermenu"></ng-content>
+  </div>
+
+  <!-- If an icon is provided, use that, otherwise use the sbb logo -->
+  <!-- TODO: make cleaner -->
+  <div #iconContent class="sbb-header-logo" [class.hidden-logo]="!iconContent.childElementCount">
```
kyubisation: Use [hidden] instead of [class.hidden-logo] and remove the css code

daniel-sc: good point, fixed this with the lastest commit

PR #210, opened 2019-11-01
feat(business): implement tabs

### LINKED ISSUE #87: feat(business): tabs styling
Design Spec: https://digital.sbb.ch/en/webapps/components/tab

DoD:
- Use business tabs in business documentation section

uzysset:
Closes #87


schematics/public2business/index.ts
```diff
@@ -6,13 +6,15 @@ export function public2business(_options: any): Rule {
     [
       'accordion',
       'autocomplete',
+      'badge',
```
kyubisation: Please also add these to .gitignore

projects/angular-showcase/src/app/features/business-component-viewer/business-component-viewer.module.ts
```diff
@@ -0,0 +1,19 @@
+import { PortalModule } from '@angular/cdk/portal';
```
kyubisation: I'm not quite content with the (almost) complete duplication of the public-component-viewer.
If possible, extract common functionality into a base class or inherit where possible from the public variant.

kyubisation:
@Frenggi Can you have a look at https://angular-staging.app.sbb.ch/210/business/components/tabs

projects/angular-showcase/src/app/features/business-component-viewer/business-component-viewer.module.ts
uzysset: @kyubisation Tried to separate it into a 'core' part. Hope that's what you were looking for 😅

kyubisation: Sorry, I might have led you astray a bit. You separated it more than I intended it to be. 😃 
But this is absolutely fine. Thank you very much.

Frenggi:
Hi @kyubisation 
The initial implementation looks ok. But the intent was to show tabs like they are known from RCP applications. Therefore the inactive tab is not as designed. Inactive != disabled. --> check attached image.
Tabs

I'm aware, that this design is a bit trickier to implement in a component than the one for tabs in public library. Therefore we might have to discuss what possibilities we have as a compromise between design and the implemented component.
The second point is the pill: there it would be great to place it inside the frame of the tab. But, I understand that you tried to reuse the component implementation from the public library. And therefore I'm not sure if it would be possible to place it inside.

uzysset:
@kyubisation Is the build pipeline currently broken?

kyubisation:
I switched from travis-ci to github actions, however there was a problem with the scripts name, which I fixed just now. You can rebase on master and then it should work.
Sorry for the slow response on this PR. I will have a meeting on monday with @Frenggi in regard to the disabled tabs. Afterwards I'll come back to this.

uzysset:
@kyubisation Any update? We'd appreciate it if we could use the tabs in our projects soon...

kyubisation:
@uzysset Please rebase the PR on top of master and force push the changes. We will merge it as it currently is and open another issue for adaptation to the new design spec.

PR #221, opened 2019-11-19
Feature/business status

verafischer:


projects/angular-showcase/src/styles.scss
```diff
@@ -141,3 +141,83 @@ nav.menu.ps .ps__rail-y {
     background-color: #fff;
   }
 }
+
+@mixin svgIconColors($fillColor, $strokeColor) {
```
verafischer: Where else to put these style definitions?

kyubisation:
@verafischer Thank you for the contribution 😃 
I'll try to review this as soon as possible.

projects/sbb-esta/angular-business/status/src/status/status.component.html
```diff
@@ -0,0 +1,20 @@
+<ng-template #content><ng-content></ng-content></ng-template>
```
kyubisation: The design specification (currently) specifies the states valid, warning, invalid (and consumer specific variants). I would like to provide these by default without having the consumer provide the icons themselves. The consumer should only provide the icons for consumer specific variants.

projects/sbb-esta/angular-business/status/src/status/status.component.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../../../angular-core/styles/common/variables';
+@import '../../../../angular-core/styles/common/colors';
+
+:host {
+  display: flex;
```
kyubisation: I would expect the status to be an inline element, so in this case flex-inline would be preferred.

projects/sbb-esta/angular-business/status/src/status/status.component.scss
```diff
@@ -0,0 +1,25 @@
...
+  flex-direction: row;
+
+  font-family: $fontSbbRoman;
+  font-size: 0.9375rem;
```
kyubisation: Feel free to use the pxToRem function. This way we can use the values provided by the design specification (which are usually in pixel), while also providing rem values for consumers.

projects/sbb-esta/angular-business/status/src/status/status.component.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { NO_ERRORS_SCHEMA } from '@angular/core';
```
kyubisation: Why is this required?

projects/sbb-esta/angular-business/status/src/status/status.component.ts
```diff
@@ -0,0 +1,31 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'sbb-status',
+  templateUrl: './status.component.html',
+  styleUrls: ['./status.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class StatusComponent {
+  /**
```
kyubisation: Type is missing: https://github.com/sbb-design-systems/sbb-angular/issues/86

projects/sbb-esta/angular-business/status/src/status/status.component.ts
```diff
@@ -0,0 +1,31 @@
...
+   * Set the status message shown on the right side.
+   */
+  @Input() message: string;
+
+  /**
+   * Set the status tooltip text. If the text is set then the tooltip appears on hover.
+   */
+  @Input() tooltipText: string;
```
kyubisation: As mentioned previously, tooltip is not part of this spec

projects/angular-showcase/src/app/business/business-examples/status-showcase/status-showcase.component.html
```diff
@@ -1,3 +1,17 @@
+<h3>Status with type</h3>
+<p>
+  <sbb-status [type]="validType" [message]="'valid type'"></sbb-status>
```
kyubisation: Here I would prefer assigning the type as an attribute, to show the simple usage
e.g. `<sbb-status type="valid" ...`

projects/angular-showcase/src/app/business/business-examples/status-showcase/status-showcase.component.html
```diff
@@ -54,18 +66,9 @@ <h3>Status Extended used in table</h3>
   <tbody>
     <tr *ngFor="let row of rows">
```
kyubisation: Having a dynamic table for an example with four entries is a bit overkill.
I'd prefer to just have the four rows.
This is also better for consumer who would like to see the source of the example.

projects/sbb-esta/angular-business/status/src/status.module.ts
```diff
@@ -12,10 +17,13 @@ import { StatusComponent } from './status/status.component';
 @NgModule({
   imports: [
     CommonModule,
-    IconCollectionModule,
     OverlayModule,
-    BrowserAnimationsModule,
-    BrowserModule
+    BrowserModule,
```
kyubisation: Remove the BrowserModule. A component module should not contain it.

projects/sbb-esta/angular-business/status/src/status/status-type.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum StatusType {
```
kyubisation: I don't think we have something like this for other modules. Is there a specific reason why you added this?

projects/sbb-esta/angular-business/status/src/status/status.component.ts
```diff
@@ -1,17 +1,28 @@
...
+import { StatusType } from './status-type.enum';
+
 @Component({
   selector: 'sbb-status',
   templateUrl: './status.component.html',
   styleUrls: ['./status.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class StatusComponent {
+  public readonly validType = StatusType.VALID;
```
kyubisation: These properties are not necessary.

projects/sbb-esta/angular-business/status/src/status/status.component.ts
```diff
@@ -1,17 +1,28 @@
...
+
   /**
    * Set the status message shown on the right side.
    */
   @Input() message: string;
 
+  /**
+   * If the type is set the icon of this status is shown
+   */
+  @Input() type: StatusType;
```
kyubisation: Use string literals as type for inputs: `'valid' | 'warning' | 'invalid'`

projects/sbb-esta/angular-business/status/status.md
```diff
@@ -2,14 +2,35 @@ The status shows the status of an object.
...
+### Status with type
 
-Set an `sbb-icon` tag as content inside the `<sbb-status>`.
+You can set a predefined type `[type]`.
+Predefined types are defined in the StatusType enum.
+At the time there are `'valid'`, `'invalid'` and `'warning'` predefined types.
 
 ```html
-<sbb-status>
+<sbb-status [type]="'valid'" [message]="'valid type'"></sbb-status>
```
kyubisation: Use the attribute variant as example here `<sbb-status type="valid" ...`

projects/sbb-esta/angular-business/styles.scss
```diff
@@ -1,3 +1,21 @@
 $sbbBusiness: true;
 
 @import '../angular-core/styles/_main.scss';
+
```
kyubisation: This should be moved to the status.component.scss, either with ::ng-deep or with ViewEncapsulation.None

efux:
Hello @kyubisation and @verafischer. We are planning on using the Tooltip component in our projects. In our use case we would like to use it without an icon and without the close icon when hovering over the element. I guess this will be fixed by this PR. Is there still some progress for this issue/PR?

kyubisation:
Migrated changes to #448

PR #251, opened 2020-01-07
feature(business): style tabs according to style guide

efux:
This PR restyles the business tabs to fit the [style guide](https://digital.sbb.ch/de/webapps/components/tab). This is a follow-up for #210.

business-tabs-example
There is no background color set for the tabs, therefore there might be some additional work necessary to have a background color to rebuild something similar to the representation in the style guide. Is this necessary/a requirement? Also I did not define any margins padding for the tab content, which is inside the borders.

image

Please note: I also had to do some changes in the example viewers of the showcase. The example viewer has already a border provided by the tab. Also the code samples leave the border of the tab content because of the negative margins. I currently left this as is, see:
image


efux:
Maybe @Frenggi can also have a look at this: https://angular-staging.app.sbb.ch/251/business/components/tabs

@kyubisation What do you think about the code samples and the margins in the business showcase? Should we split up the style sheets for the component/example viewers between public and business? I tried to avoid this for simplicity of the showcase, but it looks weird as is, especially since the API tab is also affected.

kyubisation:
I think the lack of paddings in the box seems weird. But @Frenggi has to decide that.

I would like the code samples to be contained within the borders of the tab content (reach up to the border).
I'm starting to question, whether the business part of the showcase should use the business tabs component. I wanted to directly show how the business section looks, but it seems to break the design.

Frenggi:
I'm not able to access https://angular-staging.app.sbb.ch/251/business/components/tabs. On the "latest"-build there seems to be an old visual representation.
The animated picture in the first comment above looks quite good. But a padding should be provided. Therefore I suppose 16px left (same as the tab title itself) and also 16px above.

Frenggi:
The position of the badge is not optimal. I suggest a padding top, left (to the text) and right of 2px.
4-Tabs_All_States

kyubisation:
The staging service currently has an issue with crashing unexpectedly and losing persistance. I will have a look at it soonish.

projects/angular-showcase/src/app/features/business-component-viewer/business-component-viewer/business-component-viewer.component.scss
```diff
@@ -0,0 +1,3 @@
+::ng-deep pre {
```
kyubisation: I think it would be better to do this globally in the showcase styles.scss
Or was there a reason why you wanted to avoid that?
e.g.
```scss
sbb-business {
  sbb-markdown-viewer,
  sbb-component-viewer {
    pre {
      ...
    }

    sbb-example-viewer pre {
      ...
    }
  }
}
```

efux: I did this because there is a different negative margin in business than the public showcase. What we could do is setting different tags for public and business (e.g. sbb-public-component-viewer, sbb-business-component-viewer, ..., since the tag names are otherwise irrelevant) and then configure it globally in the styles.scss. I guess that would be a cleaner solution than having multiple files and using the deprecated ng-deep.

> I'm starting to question, whether the business part of the showcase should use the business tabs component. I wanted to directly show how the business section looks, but it seems to break the design.
Do you see an easy solution to use the public tabs for the showcase and then still include the business tabs for the examples?

efux: Actually nevermind the first solution approach, the tag name is used.

kyubisation: Sorry, I'm a bit confused now. 😅 
Are you going to change the styling approach or would you like to keep it as ::ng-deep?

> Do you see an easy/clean solution to use the public tabs for the showcase and then still include the business tabs for the examples?

I think the way you now implemented it looks good, so I don't have a complaint anymore.

efux: I added a class to the different example viewers and styled the margins globally.

efux:
I formatted the styling of the badges inside the business tabs. The restructuring of the template lead to a change in the appearance of the public tabs aswell.

image

The badge is now part of the underline, when hovering over the tab or having it activated. If this is a no-go I will split up the template files of the tabs between public and business.

kyubisation:
That is unfortunately a no-go (compare with https://digital.sbb.ch/de/websites/components/tab).

One idea I had, to avoid having to restructure everything when only templates change, is extend the public2business schematic, to allow tabs.component.business.html to be copied as tabs.component.html and overwrite the original tabs.component.html at the destination directory.
What do you think?

dfsbb:
Hi everybody

I did e quick review about the new tab component for business. It looks good, thanks for the implementation.
Two open points: 
- don’t forget the border-radius of 2px in top (right and left) of the active tab. Mostly every component in the styleguide has a border-radius of 2px. So we should be consistent also in this case
- padding-right on the "li" element (tab) should be 4px instead of 2px

I hope it's not too late for these points?

Thanks and cheers

efux:
@kyubisation I adapted the public2business schematic to fit your suggestion about business and public templates. What do you think?

@dfsbb Thank you for your suggestions. As soon as the build has passed you can have a look at it again, if you want to.

kyubisation:
LGTM
Thank you for the contribution 😃 
@Frenggi and @dfsbb: Do you want to have another look?

As for merging this; This is a borderline breaking ui change, but I think we are relatively safe here. Unless there is any objection I would merge this.

dfsbb:
I did an ux-review of the mentioned points (border-radius and padding-right) and it looks good. Thanks for the implementation.

Frenggi:
For me, the tab component looks good as well. Thanks

PR #252, opened 2020-01-07
fix(textarea): trigger calculation of remaining characters correctly

### LINKED ISSUE #217: bug(textarea) remaining characters text of textarea is displayed wrong
**Describe the bug**
The TextArea component does not show the correct text (at least in my opinion) when a max length is configured and the text was not changed by the user.

image

**To Reproduce**
1. Go to https://sbb-angular.app.sbb.ch/latest/public/components/textarea
2. Add a max length
3. See text as in screenshot above

Problem also appears when directly configuring max length in the template.

**Expected behavior**
The textarea should either not show a message if there is no text present or calculate the correct number of remaining characters.

**Additional context**
The problem is apparent in the public and business part of the library.

davide1995:
This PR fixes #217 and the other issue mentioned in #217.


kyubisation:
LGTM
Thank you very much 😃

PR #270, opened 2020-02-03
docs: change name of schematics for generating icons

efux:


kyubisation:
LGTM 😃 
Thanks for catching that

PR #273, opened 2020-02-04
fix: check property pinMode before accessing

hesyar:


kyubisation:
LGTM
Thank you very much 😃

PR #279, opened 2020-02-13
test: fix tests with no expectations error

jeripeierSBB:


projects/sbb-esta/angular-public/autocomplete/src/autocomplete/autocomplete.components.spec.ts
```diff
@@ -1110,14 +1110,6 @@ describe('AutocompleteComponent', () => {
...
     });
 
-    it('should set the active item properly after filtering', fakeAsync(() => {
-      const componentInstance = fixture.componentInstance;
-
-      componentInstance.trigger.handleKeydown(downArrowEvent);
-      tick();
-      fixture.detectChanges();
-    }));
-
```
jeripeierSBB: test is duplicated by following test and has no expectations

PR #344, opened 2020-03-13
fix: correct focus of textarea and resizing behaviour

### LINKED ISSUE #194: Bug: sbb-textarea doesn't support bigger height correctly
**Describe the bug**
When sbb-textarea size is bigger than one line, only click on first line activate it.
In addition, saving such resized field and opening it results in only first line of field visible.

**Expected behavior**
Clicking anywhere in textarea activate it.
Multiple lines of text are visible after save.

**Screenshots**
textarea

jeripeierSBB:
Closes #194


projects/sbb-esta/angular-public/textarea/src/textarea/textarea.component.ts
```diff
@@ -113,26 +192,100 @@ export class TextareaComponent implements ControlValueAccessor {
...
+          this.autosize.resizeToFitContent(true);
+        });
+    });
+  }
+
+  /**
+   * @docs-private
+   */
+  ngDoCheck() {
+    this.triggerResize();
```
kyubisation: Maybe explain why this is necessary

PR #347, opened 2020-03-19
fix(business): mark chip input as invalid only when touched

jeripeierSBB:


projects/sbb-esta/angular-business/chip/src/chip-input/chip-input.component.ts
```diff
@@ -210,7 +216,8 @@ export class ChipInputComponent extends SbbChipsMixinBase
     if (value) {
       value.forEach(v => this.selectionModel.select(v));
     }
-    this._propagateChanges();
+    this._value = this.selectionModel.selected;
```
jeripeierSBB: wohl besser in if-condition reinschieben?

PR #350, opened 2020-03-23
feat(business): apply link business styling

### LINKED ISSUE #82: feat(business): link styling
Design spec: https://digital.sbb.ch/en/webapps/components/link

DoD:
- Check typography for links

vlanz:
add business styling for link component. Closes #82


projects/angular-showcase/src/app/business/business-examples/links-showcase/links.component.html
```diff
@@ -0,0 +1,57 @@
+<div class="container-fluid">
```
jeripeierSBB: Please split up the example in small examples with as less html as possible. see select showcase as an example.

projects/angular-showcase/src/app/business/business-examples/links-showcase/links.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'sbb-links-showcase',
+  templateUrl: './links.component.html'
+})
+export class LinksComponent {
```
jeripeierSBB: please rename to LinksShowcaseComponent

projects/sbb-esta/angular-business/links/src/link/link.component.ts
```diff
@@ -0,0 +1,48 @@
+/** Autogenerated by public2business schematic. DO NOT CHANGE! */
```
jeripeierSBB: Why do you commit autogenerated files? Or is it modified and the comment ist still here?

jeripeierSBB: I think you have commited the files accidentally, so please remove them :)

projects/angular-showcase/src/app/business/business-examples/links-showcase/links.component.html
jeripeierSBB: If you like, do the same in public component :)

projects/angular-showcase/src/app/business/business-examples/links-showcase/links.component.ts
vlanz: I was told not to do that https://github.com/sbb-design-systems/sbb-angular/pull/271#discussion_r375220260

projects/sbb-esta/angular-business/links/src/link/link.component.ts
vlanz: Weird how these got committed. I removed them.

projects/angular-showcase/src/app/business/business-examples/links-showcase/links.component.ts
jeripeierSBB: Bad :). We had problems when components were named the same as components from the library, so LinksComponent and LinkComponent are very similar. Generally we are updating all showcase examples and have the plan to rename them in a constant way. I started it by adding Showcase suffix again wherever i touched it. So currently it's in a mixed state :). We can leave it as it is and I will change it later.

projects/angular-showcase/src/app/business/business/business.component.ts
```diff
@@ -102,6 +105,10 @@ export class BusinessComponent implements ExampleProvider {
     dialog: {
       'dialog-showcase': new ComponentPortal(DialogShowcaseComponent)
     },
+    links: {
+      'simple-links': new ComponentPortal(SimpleLinksComponent),
```
jeripeierSBB: Because there is only one link, rename it to simple-link, and icon-link

projects/angular-showcase/src/app/public/public/public.component.ts
```diff
@@ -150,7 +151,10 @@ export class PublicComponent implements ExampleProvider {
     usermenu: { 'usermenu-showcase': new ComponentPortal(UsermenuShowcaseComponent) },
     badge: { 'badge-showcase': new ComponentPortal(BadgeShowcaseComponent) },
     button: { 'button-showcase': new ComponentPortal(ButtonShowcaseComponent) },
-    links: { 'links-showcase': new ComponentPortal(LinksShowcaseComponent) },
+    links: {
+      'simple-links': new ComponentPortal(SimpleLinksComponent),
```
jeripeierSBB: simple-link, icon-link

jeripeierSBB:
@Frenggi Can you please check the styles? https://angular-staging.app.sbb.ch/350/business/components/links Thank you :)

jeripeierSBB:
@vlanz I think social link examples in public are lost after splitting it up. Can you re-add an example? Maybe in public the simple link and icon link can be merged into one example again, what's your opinion?

Frenggi:
> @Frenggi Can you please check the styles? https://angular-staging.app.sbb.ch/350/business/components/links Thank you :)

@jeripeierSBB
- The simple link looks good. 
- The icon of the icon link should be #000 in normal status like the text itself
- The stretch mode could be provided same as it is done for public component (move the icon to the right edge of the). This we could use for menu items as well (e.g. in Header menu for suites, see https://digital.sbb.ch/en/webapps/modules/header --> 4.2 Suite)

PR #352, opened 2020-03-23
feat(business): apply file-selector business styling

### LINKED ISSUE #81: feat(business): fileselector styling
Design spec: https://digital.sbb.ch/en/webapps/components/fileselector

DoD:

vlanz:
add business styling for file-selector component, Closes #81


projects/angular-showcase/src/app/business/business-examples/file-selector-showcase/file-selector-showcase.component.html
```diff
@@ -0,0 +1,57 @@
+<article class="container-fluid">
```
jeripeierSBB: If you like, split up the examples in 3 small ones, and if you (still ;-) like, do the same in public repo.

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,28 @@
...
-  &:not(.sbb-file-selector-disabled) {
-    @include secondaryButton;
-    @include arrowbuttons;
+  @if $sbbBusiness {
+    .sbb-svgsprite-icon {
+      display: none;
+    }
 
+    &.sbb-file-selector-disabled {
+      opacity: 50%;
```
jeripeierSBB: I'd not use opacity because it depends on the background which can be non white

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,28 @@
 
 @mixin fileSelectorButton() {
```
jeripeierSBB: Without investigating further, your changes somehow messed up disabled state in public component (see https://angular.app.sbb.ch/latest/public/components/file-selector).

jeripeierSBB: your version with wrong disabled styling: https://angular-staging.app.sbb.ch/352/public/components/file-selector

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
vlanz: I took that styling from the InVision data. In the future I'll ignore opacity where it doesn't make sense and use colors instead

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
jeripeierSBB: The problem still occurs in public: the disabled state should have a border, see https://angular.app.sbb.ch/latest/public/components/file-selector for comparison. Probably` &:not(.sbb-file-selector-disabled)` is still required. for @include secondaryButton.

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
-  &:not(.sbb-file-selector-disabled) {
-    @include secondaryButton;
-    @include arrowbuttons;
+  @if $sbbBusiness {
+    .sbb-svgsprite-icon {
+      display: none;
+    }
 
+    &.sbb-file-selector-disabled {
+      background-color: $sbbColorMilk;
```
jeripeierSBB: shouldnt it be $sbbColorSilver ? (see digital.sbb.ch, invision app)

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
+      color: $sbbColorGraphite;
```
jeripeierSBB: I think it should be $sbbColorIron (see digital.sbb.ch, invision app)

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
+      &:hover {
+        background-color: $sbbColorMilk;
```
jeripeierSBB: shouldnt it be $sbbColorSilver ? (see digital.sbb.ch, invision app)

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
+        color: $sbbColorGraphite;
```
jeripeierSBB: I think it should be $sbbColorIron (see digital.sbb.ch, invision app)

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
-    @include arrowbuttons;
+  @if $sbbBusiness {
+    .sbb-svgsprite-icon {
+      display: none;
+    }
 
+    &.sbb-file-selector-disabled {
+      background-color: $sbbColorMilk;
+      color: $sbbColorGraphite;
+      &:hover {
```
jeripeierSBB: Maybe you can group &.sbb-file-selector-disabled and hover state like `&.sbb-file-selector-disabled, &.sbb-file-selector-disabled:hover`

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
```diff
@@ -11,18 +11,30 @@
...
     .sbb-svgsprite-icon {
       @include svgIconColor($buttonSecondaryColor);
     }
-  }
-  // Disabled button
-  &.sbb-file-selector-disabled {
-    @include disabledButton;
+
+    &.sbb-file-selector-disabled {
+      @include disabledButton;
```
jeripeierSBB: Only a question, wouldn't it be better to define global disabledButton style for business than defining in file selector style? so one can include it everywhere needed. @kyubisation your opinion?

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
jeripeierSBB: Very interesting, I couldn't find it there, where exactly did you find that?

vlanz: Ah well it was actually just the text that was marked that way. I must've interpreted it wrong. Nonetheless I now also made the text not opaque but the right colour.
image

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
vlanz: Interesting because it works locally on my laptop. I'll take a look.

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
vlanz: No, because it is $sbbColorIron with a 50% opacity as in the screenshot above... which I interpreted as $sbbColorGraphite

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
vlanz: No, because it is $sbbColorIron with a 50% opacity as in the screenshot above... which I interpreted as $sbbColorGraphite

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
jeripeierSBB: confusing, but thank you for your screenshot!

I think the background color is still not the right one. Compare Screenshot:
invision:
image

https://angular-staging.app.sbb.ch/352/business/components/file-selector
image

jeripeierSBB: my color picker sais #f1f1f1 which is not defined in sbb color palete... nearest color I think is $sbbColorCloud. @Frenggi Maybe you could solve our confusion on the right font and background color of the disabled file selector button in business?

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
jeripeierSBB: it can be a css bleeding issue, which depends on browser and usage. maybe that was the reason for the :not selector before

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
Frenggi: Hi @jeripeierSBB, @vlanz 
The method with opacity we chose for all variants of the button. Because we could not find colours in the SBB colour palette for disabled buttons (especially for the primary button), we decided to take the opacity way.
Because the complete design is defined for white background and we tried to support accessibility (contrasts) as far as possible, this should work that way.

--> background color: same as the enabled button
--> background opacity: 40%
--> text color: same as the enabled button
--> text opacity: 50%

projects/sbb-esta/angular-public/file-selector/src/_file-selector.scss
Frenggi: @jeripeierSBB For me, that's the correct solution. Because in design, the button of the file-selector is a normal [secondary button](https://digital.sbb.ch/en/webapps/components/button) with it's different status.

kyubisation:
LGTM 👍 
Thank you very much @vlanz and @jeripeierSBB 😃

PR #429, opened 2020-05-25
Feature/leaflet map

WyssTobi:
Add new package @sbb-esta/angular-maps-leaflet


src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+  styleUrls: ['./leaflet-map-example.component.scss'],
+})
+export class LeafletMapExampleComponent implements AfterViewChecked {
+  private _map: L.Map;
+  private _myPointLayerGroup: LayerGroup<any>;
+
+  @ViewChild('leafletMap', { static: true })
+  private _leafLetMap: LeafletMapComponent;
+
+  public options = mapOptions;
```
kyubisation, line 91: Omit public keyword

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+  public lc = layersControlConfig;
```
kyubisation, line 92: use a descriptive name here

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    for (let i = 1; i < 1000; i++) {
+      pointArray.push([this._randomNumber(45.8, 47.7), this._randomNumber(5.9, 10.5)]);
+    }
+    return pointArray;
+  }
+
+  ngAfterViewChecked() {
+    // get new mapsize when "examples"-tab is clicked
+    if (this._map) {
+      setTimeout(() => this._map.invalidateSize(), 300);
```
kyubisation, line 119: Why 300 milliseconds? And why is this not called in mapReady?

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  @HostListener('document:onchange', [])
+  onResize() {
+    console.log('asd');
+    if (this._map) {
+      this._map.invalidateSize();
+    }
+  }
+
+  public goToPoint(lat: number, lng: number, zoomlvl: number) {
```
kyubisation, line 139: public methods should be above private methods where possible

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    return Math.random() * (max - min) + min;
+  }
+
+  public mapReady(map: L.Map) {
+    this._map = map;
+  }
+
+  @HostListener('document:onchange', [])
+  onResize() {
+    console.log('asd');
```
kyubisation, line 133: This should probably be removed or rephrased

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  private _randomNumber(min, max) {
+    return Math.random() * (max - min) + min;
+  }
+
+  public mapReady(map: L.Map) {
+    this._map = map;
+  }
+
+  @HostListener('document:onchange', [])
```
kyubisation, line 131: Why onchange?
And maybe resize tracking should be done in the component. See textarea.component.ts:ngAfterViewInit for an example.

src/showcase/rollup.config.js
```diff
@@ -1,20 +1,24 @@
...
+      )
     }),
     node({
       mainFields: ['es2015', 'module', 'browser', 'jsnext:main', 'main'],
-      extensions: ['.mjs'],
+      extensions: ['.mjs']
     }),
     commonjs(),
-  ],
+    json()
```
kyubisation, line 22: Why is json necessary?

tools/leaflet/leaflet_amd_loader_config.js
```diff
@@ -0,0 +1,59 @@
+(function() {
```
kyubisation, line 1: Why is this necessary?

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
WyssTobi: This is not used anymore. I removed it.

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
WyssTobi: This workaround is necessary because the leaflet map gets its size when it is added to the DOM. At this point it is not yet displayed, but only when the user clicks the "Examples" tab. This leads to not enough tiles being loaded when the map is actually displayed. All required tiles will not be loaded until you call `invalidateSize()`. So this is not a problem of the Leaflet-map-component itself, but only in the showcase.

Is there any other way when this method can be called? For example can I get a notification when the user switches to the Examples tab?

Translated with www.DeepL.com/Translator (free version)

src/showcase/rollup.config.js
WyssTobi: When I first ran `yarn build:showcase` i ran into an issue, that it could not read the esri-leaflet plugin until I add the json plugin.
Have a look at the error message:
`[!] Error: Unexpected token (Note that you need rollup-plugin-json to import JSON files)
node_modules/esri-leaflet/package.json (2:8)`

is there another way on how I can get this working?

tools/leaflet/leaflet_amd_loader_config.js
WyssTobi: Otherwise leaflet can not be found when running tests. Is there another possible solution for that?

WyssTobi: I found a solution, sorry. I forgot to add the dependency correctly. This is not necessary indeed.

WyssTobi:
Hey @kyubisation 
Thanks for your review!
To answer your questions:
- @ajoos and I discussed this. Our thoughts were, that we will call it `@sbb-esta/angular-maps-leaflet` to be clear on the first look, that it belongs in some way to `@sbb-esta/angular-maps` and all the components there are maps and utilities for maps.
- Yes, I think there will be more components soon in this package, for example further controls. Additionaly leaflet is easy extensible by plugins, for this reason, there could be many different components coming soon.

src/showcase/rollup.config.js
kyubisation: No, it is definitely required if esri-leaflet internally imports json. Could you add a comment that esri-leaflet requires the json plugin?

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
kyubisation: I see. No, not easily.
Does it also work if you use `Promise.resolve().then(() => this._map.invalidateSize())`? We generally use Promise.resolve if something should be invoked in the next "tick".

src/showcase/rollup.config.js
WyssTobi: Esri-leaflet is only used in the showcase to add some ArcGIS (ESRI) services on the map. I could also remove it and use different data, if you think it blows the showcase up too much. 
I just used `esri-leaflet` to show how easily one can integrate ArcGIS services in a leaflet map. Otherwise I will add the comment.

kyubisation: I think it provides a good example on how to integrate the service we use at SBB. So I would say don't remove esri-leaflet, but add a comment in the rollup.config.js to indicate why we need the json plugin.

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
WyssTobi: I tested it out - I think it doesn't work. When I use `Promise.resolve().then(() => this._map.invalidateSize())` the app does not respond anymore. (But the code in the callback method is still executed.)
When I do not run it with a timeout, the map loads the tile only if the user integrates with the app.

src/angular-maps-leaflet/leaflet-map/leaflet-map/leaflet-map.component.ts
```diff
@@ -0,0 +1,150 @@
...
+      zoom ? zoom : DEFAULT_ZOOM
+    );
+  }
+
+  /** @docs-private */
+  private _registerEventHandlers() {
+    this._map.whenReady(() => {
+      this.mapReady.next(this._map);
+      // update map size when map is ready
+      setTimeout(() => this._map.invalidateSize(), 200);
```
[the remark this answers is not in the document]
WyssTobi: I tried to solve this using `Promise.resolve().then(() => this._map.invalidateSize());` but this delay is too small. If I use a delay smaller than ~100 the map won't always work in Firefox and Safari.

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
kyubisation: Ok, so it won't work with a microtask. One more thing: Does it work with just `setTimeout(() => this._map.invalidateSize())`?
Either way it would probably be a good thing to document the use case in detail in the code, so consumers can understand why it is necessary.

WyssTobi: yep, this works! 😃 thanks for the tip!

kyubisation: `Promise.resolve().then` schedules a microtask and `setTimeout` schedules a macrotask. Usually the difference doesn't matter too much, but sometimes it is relevent for waiting for other "work" on the thread.

src/showcase/app/maps-leaflet/maps-leaflet-examples/leaflet-map-examples/leaflet-map-example/leaflet-map-example.component.ts
```diff
@@ -165,4 +124,44 @@ export class LeafletMapExampleComponent implements AfterViewChecked {
...
+  /**
+   * The map gets it's size, when it is added to the DOM.
+   * At this point it is not yet displayed, but only when the user clicks the "examples"-tab.
+   * This leads to not enough tiles being loaded when the map is actually displayed, which turns out in a grey map.
+   * All required tiles will not be loaded until you call `map.invalidateSize()`.
+   * Because it is not possible to get notified when the user clicks the "examples"-tab,
+   * we use ngAfterviewChecked to get the new mapsize and load the according tiles.
+   */
+  ngAfterViewChecked() {
+    if (this._map) {
```
kyubisation, line 137: I noticed that when I am on the example page, navigate to another page (e.g. Use esri-leaflet) and navigate back to the example, it will only be a grey map.

WyssTobi: I fixed this today and it should work now.

WyssTobi:
@kyubisation I fixed now all the things you've mentioned - just in time 😃

kyubisation:
As discussed with @ajoos, we're closing this PR for now.

PR #451, opened 2020-06-09
docs: autocomplete example in table filtering

jeripeierSBB:


src/showcase/app/business/business-examples/table-examples/filter-sort-paginator-table-example/filter-sort-paginator-table-example.component.ts
```diff
@@ -47,6 +47,8 @@ export class FilterSortPaginatorTableExampleComponent implements AfterViewInit,
     description: new FormControl(''),
   });
 
+  descriptions: string[];
```
kyubisation, line 50: Use Observable<string[]> and async pipe in template

PR #457, opened 2020-06-15
feat(tabs): allow tabs to be lazy rendered by sbbTabContent directive

### LINKED ISSUE #456: Feature Request: Lazy rendering tabs
**Maybe this is rather a feature request than a bug.**

**Describe the bug**
sbb-tabs component somehow initializes the component of a tab even though the tab content does not have to be rendered.

**To Reproduce**
1. Go to https://stackblitz.com/edit/angular-sbb-vuyz1p
2. Open Console
3. OnInit lifecycle of both components has been called

However, when you check the DOM, the element app-def is not there at all. When switching between tabs, the component is only initialized once.

**Expected behavior**
OnInit method of component is only called, when the component has to be rendered. In this case this should happen, when I switch the tab.

The expected behaviour would be similar to the GhiComponent in the example. The OnInit method is called every time the state is toggled.

The default Material tabs component behaves the same way, however, there we can use the matTabContent directive for this behaviour. See https://stackblitz.com/edit/angular-9-material-starter-fuuu1a.

**Additional context**
This affects business and public.

efux:
Closes #456

Please have a look at the implementation code. If it is okay by you, I will add the necessary documentation for public and business before the merge. I used an implementation similar to the expansion panel and the tabs by Material.

### DIFF
--- a/src/angular-public/tabs/tab/tab.component.ts
+++ b/src/angular-public/tabs/tab/tab.component.ts
@@ -1,8 +1,10 @@
 import { BooleanInput, coerceBooleanProperty } from '@angular/cdk/coercion';
+import { TemplatePortal } from '@angular/cdk/portal';
 import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
+  ContentChild,
   EventEmitter,
   HostBinding,
   Input,
@@ -10,8 +12,12 @@ import {
   OnDestroy,
   OnInit,
   Output,
+  TemplateRef,
+  ViewContainerRef,
 } from '@angular/core';
 import { Subject } from 'rxjs';
+
+import { TabContent } from './tab-content';
 
 let counter = 0;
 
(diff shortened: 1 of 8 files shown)


src/angular-public/tabs/tab/tab.component.ts
```diff
@@ -92,11 +97,23 @@ export class TabComponent implements OnInit, OnChanges, OnDestroy {
    * Event generated if a tab is removed
    */
   @Output() removeChange = new EventEmitter();
+  /**
+   * Template provided in the tab content, which is lazily rendered
+   */
+  @ContentChild(TabContent) _lazyTabContent: TabContent;
```
kyubisation, line 103:
```suggestion
  @ContentChild(TabContent, { read: TemplateRef, static: true }) _lazyTabContent: TemplateRef<any>;
```

src/angular-public/tabs/tab/tab.component.ts
```diff
@@ -113,6 +130,12 @@ export class TabComponent implements OnInit, OnChanges, OnDestroy {
 
   tabMarkForCheck() {
     this._changeDetector.markForCheck();
+
+    if (this.active && this._lazyTabContent) {
```
kyubisation, line 134: This should probably be before `this._changeDetector.markForCheck()` in order to avoid timing issues.

src/angular-public/tabs/tab/tab.component.ts
```diff
@@ -92,11 +97,23 @@ export class TabComponent implements OnInit, OnChanges, OnDestroy {
...
   @Output() removeChange = new EventEmitter();
+  /**
+   * Template provided in the tab content, which is lazily rendered
+   */
+  @ContentChild(TabContent) _lazyTabContent: TabContent;
+
+  /**
+   * Portal holding the user's lazy tab content
+   */
+  portal: TemplatePortal | null = null;
```
kyubisation, line 108: Rename to _contentPortal and add @docs-private to comment

efux:
I applied your suggestions.

https://github.com/sbb-design-systems/sbb-angular/compare/55e0778e42bee200c243985702b9f0399a57d19c..d5ec160ee992495c0e703951dbc984ca0309088a

PR #539, opened 2020-07-22
refactor: migrate file-selector to new icons

jeripeierSBB:


src/angular-public/file-selector/file-selector/file-selector.component.scss
```diff
@@ -1,17 +1,15 @@
 @import '../../../angular-core/styles/common';
 
 .sbb-file-selector {
+  display: block;
```
kyubisation, line 4:
```suggestion
  display: inline-block;
```

PR #547, opened 2020-07-23
refactor: icon processflow

kyubisation:


src/angular-public/processflow/processflow/processflow.component.ts
```diff
@@ -21,8 +20,4 @@ export class ProcessflowComponent extends ProcessflowBase<ProcessflowStepCompone
   /** @docs-private */
   @ContentChildren(ProcessflowStepComponent)
   steps: QueryList<ProcessflowStepComponent>;
-
-  constructor(changeDetectorRef: ChangeDetectorRef) {
```
jeripeierSBB, line 25: does this really work? ;-)

kyubisation: Should now work with the `@Directive()` decorator on the base class.

PR #550, opened 2020-07-24
refactor: new icons for search

jeripeierSBB:


src/angular-public/search/search/search.component.ts
```diff
@@ -186,7 +199,6 @@ export class SearchComponent implements ControlValueAccessor, OnDestroy, AfterVi
   }
   private _hideSearch = true;
 
-  @HostBinding('attr.role')
   get role() {
```
kyubisation, line 202: This should probably also be inlined in host and deprecated

src/angular-public/search/search/search.component.ts
```diff
@@ -241,7 +253,6 @@ export class SearchComponent implements ControlValueAccessor, OnDestroy, AfterVi
    */
   // tslint:disable-next-line:no-input-rename
```
kyubisation, line 254: This tslint:disable can be removed

PR #557, opened 2020-07-30
feat: add mode inline to sbb-loading

jeripeierSBB:


kyubisation:
I would prefer an `@Input()` inline with coerceBooleanValue, instead of the breaking change.

src/angular-public/loading/loading/loading.component.scss
```diff
@@ -97,13 +83,15 @@ $loadingRectangleOpacities: (1, 0.5, 0), (2, 1, 0.5), (3, 0.5, 1), (4, 0.25, 0.5
     }
   }
 
-  &[inline] {
```
jeripeierSBB, line 100: It would probably better to remain definitions for [inline] to achieve backwards compability? @kyubisation

kyubisation (approved):
LGTM 👍

PR #593, opened 2020-09-08
feat(stackblitz): allow opening examples in new Stackblitz project

vlanz:
@kyubisation this is an initial draft of what the implementation of the Stackblitz example opener could look like. Before trying to tidy up the rough cut I'd like to get your opinion, especially on the following points:

- Most differences between this implementation and the Material Stackblitz opener lie in the StackblitzWriterService in the constructStackBlitzForm() function. I had to add a few lines of logic to be able to display examples with different file names / component names in the new Stackblitz project. I'm not particularly happy with the way this is constructed because it 1) creates linting errors through the "replacement placeholders" in the module files and 2) feels a bit sketchy as a whole. I'd be very much open for other ideas in this area.

- I set the dependency versions to be loaded by Stackblitz respectively to the latest available. If I knew a way, I would have set these to be the exact same as in the project, but I'm not sure how / if it is relevant.

- The solution relies on the fact that the file name uses the same wording as the Component name. Through some regexing and replacing the Component name is actually constructed from the file name provided. Again, I couldn't find a better way to define the Component name, which is needed to declare and bootstrap the chosen example in the app.module.

Feel free to comment as much as you like ;)

### DIFF
--- a/src/showcase/app/business/business-example-viewer/business-example-viewer.component.html
+++ b/src/showcase/app/business/business-example-viewer/business-example-viewer.component.html
@@ -13,6 +13,18 @@
       ></path>
     </svg>
   </button>
+
+  <button
+    type="button"
+    class="stackblitz-button"
+    (click)="openStackblitz()"
+    [disabled]="isStackblitzDisabled"
+    [attr.aria-label]="
+      isStackblitzDisabled ? 'Building StackBlitz example...' : 'Edit this example in StackBlitz'
+    "
+  >
+    <sbb-icon svgIcon="kom:link-external-small"></sbb-icon>
+  </button>
 </div>
 
 <sbb-tabs [hidden]="!showSource" class="sbbsc-tabs">
--- a/src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
+++ b/src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
@@ -1,10 +1,14 @@
 import { ComponentPortal } from '@angular/cdk/portal';
 import { Component, Input, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
-import { combineLatest, Observable } from 'rxjs';
+import { combineLatest, Observable, zip } from 'rxjs';
 import { map } from 'rxjs/operators';
 
 import { HtmlLoader } from '../../html-loader.service';
+import {
+  ExampleData,
+  StackblitzWriterService,
+} from '../stackblitz-writer/stackblitz-writer.service';
 
 @Component({
   selector: 'sbb-example-viewer',
@@ -19,6 +23,10 @@ export class ExampleViewerComponent implements OnInit {
   scss: Observable<string>;
   showSource = false;
   title: Observable<string>;
+
+  isStackblitzDisabled = true;
+  stackBlitzForm: HTMLFormElement;
+
   get label() {
     return this.name
       .replace(/-/g, ' ')
@@ -26,7 +34,18 @@ export class ExampleViewerComponent implements OnInit {
       .replace(' Showcase', '');
   }
 
-  constructor(private _htmlLoader: HtmlLoader, private _route: ActivatedRoute) {}
+  get componentName() {
+    return this.name
+      .replace(/(^[a-z]|-[a-z])/g, (m) => m.toUpperCase())
+      .replace(/-/g, '')
+      .concat('Component');
+  }
+
+  constructor(
+    private _htmlLoader: HtmlLoader,
+    private _route: ActivatedRoute,
+    private _stackBlitzWriter: StackblitzWriterService
+  ) {}
 
   ngOnInit(): void {
     this.title = combineLatest([this._route.params, this._route.data]).pipe(
@@ -43,5 +62,43 @@ export class ExampleViewerComponent implements OnInit {
     this.ts = this._htmlLoader.with(this._route).fromExamples(this.name, 'ts').observe();
 
     this.scss = this._htmlLoader.with(this._route).fromExamples(this.name, 'scss').observe();
+
+    const exampleContents = ['ts', 'html', 'scss'].map((type: 'ts' | 'html' | 'scss') =>
+      this._htmlLoader
+        .with(this._route)
+        .fromSourceExamples(this.name, type)
+        .observe()
+        .pipe(map((content) => ({ name: `${this.name}.${type}`, content })))
+    );
+    zip(...exampleContents).subscribe((results) =>
+      this._createStackblitzForm(results.filter((result) => !!result.content))
+    );
+  }
+
+  private _createStackblitzForm(files: { name: string; content: string }[]) {
+    const example = new ExampleData({
+      componentName: this.componentName,
+      selectorName: `${this.name}`,
+      description: this.label,
+      indexFilename: files[0].name,
+      exampleFiles: files,
+      business: this._route.snapshot.data.library === 'business',
+    });
+    this._stackBlitzWriter
+      .constructStackBlitzForm(example)
+      .then((stackBlitzForm: HTMLFormElement) => {
+        this.stackBlitzForm = stackBlitzForm;
+        this.isStackblitzDisabled = false;
+      });
+  }
+
+  openStackblitz() {
+    // When the form is submitted, it must be in the document body. The standard of forms is not
+    // to submit if it is detached from the document. See the following chromium commit for
+    // more details:
+    // https://chromium.googlesource.com/chromium/src/+/962c2a22ddc474255c776aefc7abeba00edc7470%5E!
+    document.body.appendChild(this.stackBlitzForm);
+    this.stackBlitzForm.submit();
+    document.body.removeChild(this.stackBlitzForm);
   }
 }
--- a/src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
+++ b/src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
@@ -1,0 +1,218 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+
+export class ExampleData {
+  indexFilename: string;
+  description: string;
+  selectorName: string;
+  componentName: string;
+  exampleFiles: { name: string; content: string }[];
+  business: boolean;
+
+  constructor(data: { [P in keyof ExampleData]: ExampleData[P] }) {
+    this.indexFilename = data.indexFilename;
+    this.description = data.description;
+    this.selectorName = data.selectorName;
+    this.componentName = data.componentName;
+    this.exampleFiles = data.exampleFiles;
+    this.business = data.business;
+  }
+}
+
+const STACKBLITZ_URL = 'https://run.stackblitz.com/api/angular/v1';
+
+/**
+ * Path that refers to the docs-content from the "@angular/components-examples" package.
+ */
+const CONTENT_PATH = 'assets/docs-content/examples-source/';
+
+/**
+ * Path that refers to the generic project template files.
+ */
+const TEMPLATE_PATH = 'assets/stackblitz/';
+
+/**
+ * Filenames of the generic project template files.
+ */
+const TEMPLATE_FILES = [
+  'angular.json',
+  'tsconfig.json',
+  'src/main.ts',
+  'src/polyfills.ts',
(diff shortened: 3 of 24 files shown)


kyubisation:
@vlanz Thank you very much 👍 
I will try to have a detailed look at it today

src/showcase/app/business/business-example-viewer/business-example-viewer.component.html
```diff
@@ -13,6 +13,27 @@ <h3>{{ label }}</h3>
...
+
+  <button
+    type="button"
+    (click)="openStackblitz()"
+    [disabled]="isStackblitzDisabled"
+    [attr.aria-label]="
+      isStackblitzDisabled ? 'Building StackBlitz example...' : 'Edit this example in StackBlitz'
+    "
+  >
+    <svg
```
kyubisation, line 25: Please use `<sbb-icon svgIcon="kom:link-external-small"></sbb-icon>`

src/showcase/app/business/business-example-viewer/business-example-viewer.component.ts
```diff
@@ -7,4 +7,9 @@ import { ExampleViewerComponent } from '../../shared/component-viewer/example-vi
   templateUrl: './business-example-viewer.component.html',
   styleUrls: ['./business-example-viewer.component.css'],
 })
-export class BusinessExampleViewerComponent extends ExampleViewerComponent {}
+export class BusinessExampleViewerComponent extends ExampleViewerComponent implements OnInit {
+  ngOnInit() {
+    this.business = true;
```
kyubisation, line 12: As discussed, use the library key from route data in loader-builder.ts

src/showcase/assets/stackblitz/src/app/sbb.module.ts
```diff
@@ -0,0 +1,76 @@
+import { NgModule } from '@angular/core';
+import {
```
kyubisation, line 2: As discussed with bazel rule:
genrule used-icons in src/BUILD.bazel
genrule used-icons in src\angular-core\schematics\BUILD.bazel

Modules in:
- src\angular-business\config.bzl
- src\angular-public\config.bzl

kyubisation (changes requested):
Looks very good already 👍

src/showcase/assets/stackblitz/src/app/sbb.module.ts
vlanz: I might need more bazel knowledge to do this, I seem to get stuck on very basic problems e.g. how to test my changed bazel BUILD files... if you want I can invest some time into learning more about bazel, otherwise I'll leave this one up to you ;)

src/showcase/app/shared/loader-builder.ts
```diff
@@ -8,6 +8,10 @@ export class LoaderBuilder {
 
   constructor(private _http: HttpClient, private _route: ActivatedRoute) {}
 
+  get library() {
```
kyubisation, line 11: This could probably be moved into the example viewer component

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -43,5 +62,70 @@ export class ExampleViewerComponent implements OnInit {
     this.ts = this._htmlLoader.with(this._route).fromExamples(this.name, 'ts').observe();
 
     this.scss = this._htmlLoader.with(this._route).fromExamples(this.name, 'scss').observe();
+
+    this._htmlLoader
```
kyubisation, line 66: This could be refactored to:
```ts
    const exampleContents = ['ts', 'html', 'scss'].map((type: 'ts' | 'html' | 'scss') =>
      this._htmlLoader
        .with(this._route)
        .fromSourceExamples(this.name, type)
        .observe()
        .pipe(map((content) => ({ name: `${this.name}${type}`, content })))
    );
    forkJoin(exampleContents).subscribe((result) => this.createStackblitzForm(result));
```

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -43,5 +62,70 @@ export class ExampleViewerComponent implements OnInit {
...
+                    })
+                  );
+              })
+            );
+        })
+      )
+      .subscribe((result) => this.createStackblitzForm(result));
+  }
+
+  createStackblitzForm(files) {
```
kyubisation, line 101: This could be made private (I think)

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -19,14 +23,29 @@ export class ExampleViewerComponent implements OnInit {
...
+
   get label() {
     return this.name
       .replace(/-/g, ' ')
       .replace(/(^[a-z]| [a-z])/g, (m) => m.toUpperCase())
       .replace(' Showcase', '');
   }
 
-  constructor(private _htmlLoader: HtmlLoader, private _route: ActivatedRoute) {}
+  get componentName() {
```
kyubisation, line 37: This could be inlined

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -43,5 +62,70 @@ export class ExampleViewerComponent implements OnInit {
...
+  createStackblitzForm(files) {
+    const example = new ExampleData({
+      componentName: this.componentName,
+      selectorName: `${this.name}`,
+      description: this.label,
+      indexFilename: files[0],
+      exampleFiles: files,
+      business: this._htmlLoader.with(this._route).library === 'business',
+    });
+    if (example) {
```
kyubisation, line 110: This if check is not needed, since example is always true

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export class ExampleData {
+  indexFilename: string;
+  description: string;
+  selectorName: string;
+  componentName: string;
+  exampleFiles: { name: string; content: string }[];
+  business: boolean;
+
+  constructor(data: any) {
```
kyubisation, line 13: Replace any with `{ [P in keyof ExampleData]: ExampleData[P] }`

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+        '',
+        false
+      );
+
+      TEMPLATE_FILES.forEach((file) =>
+        this._readFile(file, TEMPLATE_PATH).subscribe(
+          (response: string) => {
+            if (file.includes('app.module.ts')) {
+              // add currently selected component to declarations and bootstrap it
+              response = response.replace(/\{0\}/g, data.componentName);
```
kyubisation, line 128: I think it would be better to use the corresponding name as the token key: e.g. {componentName} instead of {0}

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+            this._addFileToForm(form, data, response, file, TEMPLATE_PATH, false);
+          },
+          (error) => console.log(error)
+        )
+      );
+
+      // remove ".component" filename section from html template URL
+      data.exampleFiles.some((file) => {
+        if (file.name.includes('.ts')) {
+          file.content = file.content.replace('component.html', 'html');
```
kyubisation, line 146: Since we compile scss with bazel, we have .css endings in styleUrl, but the source files end in .scss
Also replace `.css` with `.scss`

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+              );
+            }
+            this._addFileToForm(form, data, response, file, TEMPLATE_PATH, false);
+          },
+          (error) => console.log(error)
+        )
+      );
+
+      // remove ".component" filename section from html template URL
+      data.exampleFiles.some((file) => {
```
kyubisation, line 144: Replace .some with .forEach here

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+        if (file.name.includes('.ts')) {
+          file.content = file.content.replace('component.html', 'html');
+        }
+      });
+
+      data.exampleFiles.map((file) =>
```
kyubisation, line 150: Replace .map with .forEach here

src/showcase/assets/stackblitz/src/app/app.module.ts
```diff
@@ -0,0 +1,14 @@
+import { NgModule } from '@angular/core';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { BrowserModule } from '@angular/platform-browser';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+
+import { SbbModule } from './sbb.module';
+import { {0} } from './{1}';
+
+@NgModule({
+  imports:      [ BrowserModule, BrowserAnimationsModule, FormsModule, ReactiveFormsModule, SbbModule ],
```
kyubisation, line 10: Add the SbbIconModule from '@sbb-esta/angular-core/icon' and also the provider SBB_ICON_REGISTRY_PROVIDER (See the showcase app.module.ts for comparison)

src/showcase/assets/stackblitz/src/app/sbb.module.ts
kyubisation: Maybe it would be easier to create two separate files for public and business and adapt the import path in app.module.ts depending on public or business.

I just checked how to  pass variables to a genrule and it is not simple: https://docs.bazel.build/versions/2.0.0/be/make-variables.html#custom_variables

If you want, you can investigate how to do this in bazel, but you don't have to.

kyubisation (changes requested):
This time a more in depth review

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -43,5 +62,70 @@ export class ExampleViewerComponent implements OnInit {
...
+            .fromSourceExamples(this.name, 'html')
+            .observe()
+            .pipe(
+              mergeMap((html) => {
+                return this._htmlLoader
+                  .with(this._route)
+                  .fromSourceExamples(this.name, 'scss')
+                  .observe()
+                  .pipe(
+                    map((scss) => {
```
jeripeierSBB, line 83: I think you need to check if a scss file really exists. In your current solution, if there is no scss file for an example, a fallback html page is loaded and results in a corrupt stackblitz example. (e.g. https://angular-staging.app.sbb.ch/593/public/components/autocomplete)

src/showcase/app/shared/component-viewer/stackblitz-writer/stackblitz-writer.service.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Path that refers to the docs-content from the "@angular/components-examples" package.
+ */
+const CONTENT_PATH = 'assets/docs-content/examples-source/';
+
+/**
+ * Path that refers to the generic project template files.
+ */
+const TEMPLATE_PATH = '/src/showcase/assets/stackblitz/';
```
kyubisation, line 33: This should be `assets/stackblitz` to be compliant with the base href.
This probably relates to the comment from @jeripeierSBB.

vlanz:
@kyubisation I added the previously discussed genrules to load modules in Stackblitz based on the current business/public library context. I had to extend the code a little to allow it to run smoothly on MacOs as well, since the MacOs version of sed does not support certain escape sequences. This, though, requires all MacOs developers to install [gnu-sed](http://www.gnu.org/software/sed/) (brew install gnu-sed) on their machine.
I also had to remove the newline escape sequence from the join operation, as I couldn't find any way to correctly print a new line on MacOs.

kyubisation (approved):
@vlanz Sorry for the delay in merging this. Thank you very much for your work!
I believe this will help immensely in supporting consumers and bug reports/tracking.

PR #606, opened 2020-09-22
fix(core): fix sbb-icon using more space than icon

kyubisation:


jeripeier (approved):
Basically LGTM, but can have some influence in our own usage, did you check that?

PR #615, opened 2020-09-28
fix(core/business-date-adapter): use year pivot to determine century of 2-digit year

### LINKED ISSUE #613: Feature request: Business Date Adapter supported formats
Currently the BUSINESS_DATE_ADAPTER parses dates like `11112020` to 11.11.2020, which is correct. However, when I enter 111199 it returns 11.11.2099, which seems odd.

When I enter `11.11.99` it parses correctly to 11.11.1999. I would like to receive the same date, when entering `111199`. They should react both the same, depending on the configured year pivot.

Stackblitz for testing: https://stackblitz.com/edit/angular-sbb-kecn5f?file=src%2Fapp%2Fapp.module.ts

efux:
This should fix the issue mentioned in #613. From my point of view, there is no reason for the additional logic and it should be handled the same as when using the NativeDateAdapter.

Closes #613


kyubisation (approved):
LGTM 👍 
Thanks for the fix!

PR #626, opened 2020-10-21
fix(business): correct tooltip colors

### LINKED ISSUE #604: Tooltip icon with inherited background color
Currently, the icon of the tooltip has background color white (https://github.com/sbb-design-systems/sbb-angular/blob/c3d743091cb132c27f7deb920ef3543822874442/src/angular-business/tooltip/tooltip/tooltip.component.scss#L37). Which looks odd on other background colors than white.

What about setting background color to inherited?

jeripeierSBB:
Closes #604


kyubisation (approved):
LGTM 👍

PR #630, opened 2020-10-28
fix: change tooltip trigger css display from block to inline-block

jeripeierSBB:
BREAKING CHANGE: This commit changes the css display property of the tooltip trigger
from block to inline-block.
If you still like to use the block variant, use the following css definition: 
`.sbb-tooltip-trigger { display: block; }`.

### DIFF
--- a/src/angular-business/tooltip/tooltip/tooltip.component.scss
+++ b/src/angular-business/tooltip/tooltip/tooltip.component.scss
@@ -3,30 +3,20 @@ $sbbBusiness: true;
 @import '../../../angular-core/styles/common';
 
 $tooltipBorderRadius: 2;
-$tooltipBorderWidth: 2;
+$tooltipBorderWidth: if($sbbBusiness, 1, 2);
 $tooltipIconsBorderWidth: 1;
 $tooltipMaxWidth: 460;
 $tooltipIconsSize: 24;
-$tooltipContentPadding: 24;
-$tooltipColor: $sbbColorGrey;
-$tooltipBorderColor: $sbbColorGrey;
-$tooltipActiveColor: $sbbColorRed;
+$tooltipContentPadding: if($sbbBusiness, 8, 24);
+$tooltipColor: if($sbbBusiness, $sbbColorBlack, $sbbColorGrey);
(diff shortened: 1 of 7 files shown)


src/angular-business/tooltip/tooltip/tooltip.component.scss
```diff
@@ -3,31 +3,21 @@ $sbbBusiness: true;
...
-  $tooltipContentPadding: 8;
-  $tooltipBorderWidth: 1;
-  $tooltipShadowSize: 4;
-  $tooltipShadowBorderWidth: 14;
-  $tooltipColor: $sbbColorBlack;
-  $tooltipBorderColor: $sbbColorIron;
-  $tooltipActiveColor: $sbbColorRed125;
-}
-
 @mixin tooltipIcon {
```
kyubisation, line 21: Would it be possible to refactor this to simply be a css class?

src/angular-business/tooltip/tooltip/tooltip.component.scss
```diff
@@ -3,31 +3,21 @@ $sbbBusiness: true;
 @import '../../../angular-core/styles/common';
 
 $tooltipBorderRadius: 2;
```
kyubisation, line 5: Could any of these be inlined?

src/angular-business/tooltip/tooltip/tooltip.component.scss
```diff
@@ -183,76 +170,80 @@ sbb-tooltip {
...
+      padding-bottom: toPx($tooltipContentPadding * $scalingFactor5k);
+      padding-right: toPx(
+        $tooltipContentPadding * $scalingFactor5k + $tooltipIconsSize + $tooltipContentPadding *
+          $scalingFactor5k
+      );
     }
   }
 }
 
+.sbb-tooltip-trigger-hover {
```
kyubisation, line 237: Move these inside their specification as:
```css
.sbb-tooltip-close-icon {
  ...

  .sbb-tooltip-trigger-hover & {
    ...
  }
}
```

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #632, opened 2020-10-29
fix: allow tooltip usage inside sbb-label

jeripeierSBB:
in IE only working with double click :)


src/angular-public/field/label/label.component.scss
```diff
@@ -14,10 +17,16 @@
...
-  }
-
   @include label;
+
+  .sbb-tooltip-icon {
+    &:not(:hover):not(:focus) {
+      color: inherit;
+    }
+
+    width: 16px;
```
kyubisation, line 27: What about 4k/5k variants for public?

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #636, opened 2020-11-02
refactor: replace async/await with plain promises

kyubisation:
This is due to Zone.js not working with async/await: https://github.com/angular/angular/issues/31730


jeripeierSBB (approved):
LGTM 👍

PR #655, opened 2020-11-27
fix(business): fix link stretch mode and adapt css classes

jeripeierSBB:
To follow our css naming guide lines we had to rename the following css class of sbb-link-business.

BREAKING CHANGE: css class `.sbb-link-normal` was renamed to `.sbb-link-business-normal`.
css class `.sbb-link-stretch` was renamed to `.sbb-link-business-stretch`.


kyubisation (approved):
LGTM 👍

PR #673, opened 2020-12-02
fix: add warning in migration for form-field control requirement

kyubisation:
sbb-form-field now requires a compatible form field control. In the migration we add
a warning, that this is now required and where in the consumers code this is a problem.


jeripeierSBB (approved):
LGTM 👍

PR #682, opened 2020-12-08
fix: allow sticky elements in expansion panel content

### LINKED ISSUE #620: Bug: overflow on expansion panel breaks sticky elements
**Describe the bug**
We have a sticky element inside an expansion-panel that we need to position 'sticky'. This fails as the expansion panel is (falsly!) detected as nearest scrolling ancestor, because it has `overflow: hidden`.

An analogous problem was already fixed for the process flow: #399

**Expected behavior**
The sticky element should be sticky based on the actual scrolling ancestor.

 - OS: Win10
 - Browser Chrome 86
 - Version SACL 10.3.0

jeripeierSBB:
Closes #620

### DIFF
--- a/src/angular-public/accordion/expansion-panel/expansion-panel.component.scss
+++ b/src/angular-public/accordion/expansion-panel/expansion-panel.component.scss
@@ -4,7 +4,6 @@
   box-sizing: content-box;
   display: block;
   margin: 0;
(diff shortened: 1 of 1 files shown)


kyubisation (approved):
LGTM 👍

PR #704, opened 2021-01-26
fix: fix ribbon color of sbb-header

sueltschi-sbb:
Correct header title according to https://digital.sbb.ch/de/webapps/modules/header
* left title offset of 54px 
* header ribbon text color white

I just noticed, that the left margin is only incorrect when there is no visible hamburger menu, which happens when there is no 'sbb-app-chooser-section' and there is enough space on the screen:

OK
image

NOK
image

so fixing the 'margin-left' wont work.

I'll fix it for now by setting a default 'sbb-app-chooser-section', fixing the color would still be nice tough


kyubisation:
Thank you for the PR. We will have a look as soon as possible.

PR #706, opened 2021-01-29
refactor(angular): adapt typography for the angular package

kyubisation:


src/angular/styles/includes/_mixins.scss
```diff
@@ -335,108 +136,14 @@
   transform: translateY(-50%);
 }
 
+@mixin absoluteSize() {
```
jeripeierSBB, line 139: Do we need that?

PR #716, opened 2021-02-08
refactor(angular): select merge

jeripeierSBB:


src/angular/styles/includes/_mixins.scss
```diff
@@ -164,6 +164,53 @@
...
+@mixin horizontalBorderAndShadowAsBoxShadow($attached: below, $borderWidth: 2, $scalingFactor: 1) {
+  $sign: 1;
+  @if $attached == above {
+    $sign: -1;
+  }
+  box-shadow: boxShadow($sign, $borderWidth, $scalingFactor),
+    horizontalBorderBoxShadow(-$sign, $borderWidth, $scalingFactor);
+}
+
+@function boxShadow($sign, $borderWidth, $scalingFactor) {
```
kyubisation, line 176: Is this function necessary?

src/angular/styles/includes/_mixins.scss
```diff
@@ -164,6 +164,53 @@
...
+  @return 0 pxToRem($sign * $borderWidth * 4 * $scalingFactor) 0 rgba(0, 0, 0, 0.15);
+}
+
+@function horizontalBorderBoxShadow($sign, $borderWidth, $scalingFactor) {
```
kyubisation, line 180: Is this function necessary?

src/angular/styles/includes/_mixins.scss
```diff
@@ -164,6 +164,53 @@
   }
 }
 
+@mixin horizontalBorderAndShadowAsBoxShadow($attached: below, $borderWidth: 2, $scalingFactor: 1) {
```
kyubisation, line 167: Is it intended to use horizontalBorderAndShadowAsBoxShadow and panelOffsetCorrection separately in isolation? If not, these could be merged into one mixin.

src/angular/styles/includes/_mixins.scss
```diff
@@ -164,6 +164,53 @@
...
+    $offset-top: $offset-top - 0.05; // better rendering e.g. in 4k mode
+  }
+  @if $offset-top > 0 {
+    $offset-top: ceil($offset-top);
+  }
+  position: relative;
+  top: pxToRem($offset-top);
+}
+
+@mixin borderLeftAsAfter($color) {
```
kyubisation, line 202: Is there a specific reason as to why use ::after instead of ::before?

kyubisation (changes requested):
Great work so far 😄 
There is one thing I mentioned via Teams in regards to property ordering, which seems broken?

src/angular/styles/_typography.scss
```diff
@@ -1371,3 +1486,356 @@ $trackColor: $sbbColorMilk;
...
+  @include panelOffsetCorrection();
+
+  hr {
+    margin: pxToRem(10) 0;
+    padding: 0;
+    border: none;
+    border-top: solid pxToRem(1) $sbbColorCloud;
+  }
+
+  @include sbbStandard() {
```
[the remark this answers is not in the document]
jeripeierSBB: Yes we can :-)

src/angular/select/select/select.component.scss
```diff
@@ -0,0 +1,97 @@
+@import '../../../angular/styles/common';
+
+.sbb-select {
+  display: flex;
+  cursor: pointer;
+
+  padding-right: pxToRem(12);
+  padding-bottom: pxToRem(10);
+
+  @include sbbStandard() {
```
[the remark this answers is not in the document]
jeripeierSBB: otherwise the styles from sbb-input-element would be applied. I added the following comment:
 // Use variant mixin here to be specific enough to overwrite sbb-input-element styles

src/angular/styles/_typography.scss
```diff
@@ -353,45 +305,48 @@ h4,
...
+    line-height: pxToRem(32);
   }
 }
 
 p,
 .sbb-copy {
-  font-size: pxToRem(15);
+  font-size: pxToRem($sizeFontDefault);
   margin: 1em 0 0; // paragraph as default
+  line-height: 1.733;
```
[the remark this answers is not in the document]
jeripeierSBB: I decided to leave the line-heights, when the value was exact and will be used in other sizes as well. Otherwise we would have to add media queries for each size. What do you think?

src/angular/styles/includes/_mixins.scss
jeripeierSBB: will be used in usermenu and contextmenu

src/angular/styles/includes/_mixins.scss
jeripeierSBB: will be used in dialog

src/angular/styles/includes/_mixins.scss
jeripeierSBB: From the technically perspective it could be merged, but I thought it's better to have them separated because its different functions. What do you think?

src/angular/styles/includes/_mixins.scss
jeripeierSBB: no, changed it to before

kyubisation (approved):
LGTM 👍

PR #721, opened 2021-02-12
refactor(angular): set variant only on html tag

jeripeierSBB:


src/components-examples/angular/textarea/textarea-forms/textarea-forms-example.html
```diff
@@ -11,16 +11,13 @@ <h4>Model</h4>
...
-</p>
-<p>
-  <sbb-form-field label="Max Length">
-    <input sbbInput type="number" [(ngModel)]="maxlength" />
-  </sbb-form-field>
-</p>
+<sbb-form-field label="Min Length">
+  <input [(ngModel)]="minlength" sbbInput type="number" />
+</sbb-form-field>
+<br />
```
kyubisation, line 17: We should not use br for separation
Maybe we need to define a class similar to .sbb-checkbox-group-vertical?

kyubisation (changes requested):
Thank you for the changes 😄 
I noticed a few things

kyubisation (approved):
LGTM 👍

PR #728, opened 2021-02-18
refactor(angular): remove .component from file names

jeripeierSBB:
option renaming is missing. will be provided with select changes (material)


kyubisation (approved):
LGTM 👍

PR #729, opened 2021-02-18
refactor(angular): rename files and line up with material

jeripeierSBB:


src/angular/core/option/option-parent.ts
```diff
@@ -0,0 +1,19 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * Describes a parent component that manages a list of options.
+ * Contains properties that the options can inherit.
+ * @docs-private
+ */
+export interface SbbOptionParentComponent {
+  disableRipple?: boolean;
```
kyubisation, line 9: We don't have ripple

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #737, opened 2021-02-23
refactor: lower specificity of .sbb-icon-fit selector

kyubisation:
The current selector is fairly aggressive and requires constructs
like .sbb-icon.sbb-icon {} to overwrite reliably. This change should
address that problem partially.


jeripeierSBB (approved):
LGTM

PR #739, opened 2021-02-23
fix: remove positioning fix for tick

kyubisation:
Previously the checkbox tick was positioned slightly below specification.
This has been fixed in the icon CDN, which caused the tick to be placed
above the intended position. This change fixes the positioning.


jeripeierSBB (approved):

jeripeierSBB (approved):
LGTM

PR #744, opened 2021-03-02
fix(business): avoid Expression has changed for menus collapsed state

kyubisation:
Setting a value synchronously in ngAfterViewInit causes a ExpressionChangedAfterItHasBeenCheckedError.
Waiting a tick via Promise.resolve() solves this issue.


jeripeierSBB (approved):
LGTM

PR #746, opened 2021-03-04
fix(tabs): prevent re-creation of existing tab instances on (redundant) activation of active tab

### LINKED ISSUE #638: Bug: Tabs component lazy rendering creates new components for each input change
**Describe the bug**
The Tabs component creates a new component tree for each change of inputs of the tabs root component when lazy rendering is applied under unknown circumstances. This leads to many bugs that are difficult to find. Our components heavily rely on lifecycle hooks (e.g. [**first change**](https://angular.io/api/core/SimpleChange)). If a input change in a Tabs root component occurs, **first change** for exmple is true again since the component tree is created completly new.

I tried to reproduce the behaviour in an simple example in [stackblitz](https://stackblitz.com/edit/angular-sbb-kkezwu).
Unfortunately i was not able to reproduce the bug with a simple example. I assume the root cause is more complex, perhaps caused with complex project structures or libraries e.g. ngrx. In any case project intern we can reproduce the behaviour simply through enable lazy loading of tabs compoentent. Without lazy loading of abs component everything behaves as excpected.

**Expected behavior**
Lazy rendering of Tabs component should not create a new component tree for each input change. 
Alternatively the circumstance should be documented.

 - OS: Windows
 - Browser: all

**Additional context**
Note: We use **on push** change dectection.

daniel-sc:
This prevents problems when tabs are opened programmatically based on external information,
such as query params.

Closes #638

### DIFF
--- a/src/angular-public/tabs/tabs/tabs.component.spec.ts
+++ b/src/angular-public/tabs/tabs/tabs.component.spec.ts
@@ -130,6 +130,17 @@ describe('SbbTabs', () => {
     expect(component.numberOfTimesSubComponentHasBeenInitialized).toEqual(1);
   });
 
+  it('should not render lazy tab content again, when opening active tab', async () => {
+    component.tabsComponent.openTabByIndex(3);
+    fixture.detectChanges();
+    await fixture.whenStable();
+    component.tabsComponent.openTabByIndex(3); // open active tab (redundantly)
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍 
Thank you for the contribution!

jeripeierSBB (approved):

PR #751, opened 2021-03-10
docs(maps): specifiy esri-loader version to use

ajoos:
angular-maps component has to be used with the same esri-loader version as used on build time of
the component. any other, more recent version does not work properly. Currently, the matching
and working version is version is 2.13.0.

### DIFF
--- a/src/angular-maps/getting-started.md
+++ b/src/angular-maps/getting-started.md
@@ -29,13 +29,13 @@ You can create now your project as described in the official [Angular CLI docume
 Just after you created your own Angular project, in order to include the library, you have to install the `@sbb-esta/angular-maps`, `@types/arcgis-js-api` and `esri-loader` dependencies:
 
 ```sh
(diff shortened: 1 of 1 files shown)


SahanFaruk (approved):

jeripeierSBB (approved):
Thanks for your PR. Basically it's okay as a documentation fix but I think we should address updating the libraries and components soon, to keep up with esri-loader changes. If you like, we can discuss it with @kyubisation .

ajoos:
> Thanks for your PR. Basically it's okay as a documentation fix but I think we should address updating the libraries and components soon, to keep up with esri-loader changes. If you like, we can discuss it with @kyubisation .

I fully agree and this discussion has actually already started. The documentation fix is there to avoid further confusion of possible users of the angular-maps compontent until we agreed on a more future-proof way to load the esri libs.

PR #759, opened 2021-03-16
fix: hide content of nested accordion in collapsed state

### LINKED ISSUE #758: Bug: Visibility of nested Accordions is not set properly
**Describe the bug**
Visibility of nested accordions is not set properly. The expansion-panel-content of an open child accordion keeps visbility: visible when closing the the parent accordion. 
This can be fixed by the following workaround: 
`.sbb-expanded {
  ::ng-deep .sbb-expansion-panel-content {
    visibility: inherit !important;
  }
}`

However, the accordion should deal with nested accordions properly.

**To Reproduce**
Reproduction: https://stackblitz.com/edit/angular-1svegp
The example "Panel with lazy loaded content" at: https://sbb-angular.app.sbb.ch/public/components/accordion shows the behavior as well

jeripeierSBB:
Closes #758

### DIFF
--- a/src/angular-public/accordion/expansion-panel/expansion-panel.component.scss
+++ b/src/angular-public/accordion/expansion-panel/expansion-panel.component.scss
@@ -37,7 +37,9 @@
   display: flex;
   flex-direction: column;
   overflow: visible;
-  &.ng-animating {
+
+  &.ng-animating,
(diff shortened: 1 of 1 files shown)


kyubisation (approved):
LGTM 👍

PR #767, opened 2021-03-19
refactor: migrate checkbox panel

kyubisation:


src/components-examples/angular/checkbox-panel/checkbox-panel-group/checkbox-panel-group-example.ts
```diff
@@ -0,0 +1,77 @@
+import { Component } from '@angular/core';
```
jeripeierSBB, line 1: Maybe we should declare somewhere that this is more a showcase than an example.

jeripeierSBB (changes requested):
Basically especially from a code perspective it looks great :-)

I found some issues:
* Somehow behaviour of sbb checkbox changed (see lean/standard switch in header)
* Generally I think the mobile responsive behaviour is not correct comparing with examples on previewserver.sbb.ch
* I'd suggest to change order of examples that the "big" group example is not the first one.
* Panel img on https://angular-staging.app.sbb.ch/767-merge/angular/components/checkbox-panel/examples not working, but probably a staging issue
* Disabling a panel change its height
* Checkbox panel content example has wrong mobile behaviour
* In schematics tests time_input_input file should also be removed

jeripeierSBB (approved):
LGTM, great work! Thanks for all changes

PR #771, opened 2021-03-24
refactor(angular): show warning for variant limited components

jeripeierSBB:


src/showcase-merge/app/shared/component-viewer/one-variant-only-notification/one-variant-only-notification.component.html
```diff
@@ -0,0 +1,14 @@
+<sbb-notification
+  *ngIf="variantOnly"
+  type="info"
+  class="sbb-one-variant-only-notification"
+  readonly
+>
+  <span class="sbb-one-variant-only-notification-title">{{ variantOnly }}-only component</span>
```
kyubisation, line 7: `component` doesn't really reflect what this is. Maybe reduce it to just {{ ... }}-only? Or what do you think?

src/showcase-merge/app/shared/component-viewer/one-variant-only-notification/one-variant-only-notification.component.html
```diff
@@ -0,0 +1,14 @@
...
+  <p>
```
kyubisation, line 8: Should we maybe use textexpand to show/hide the details?

src/showcase-merge/app/shared/component-viewer/one-variant-only-notification/one-variant-only-notification.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Input } from '@angular/core';
+
+import { ShowcaseMetaEntry } from '../../meta';
+
+@Component({
+  selector: 'sbb-one-variant-only-notification',
+  templateUrl: './one-variant-only-notification.component.html',
+  styleUrls: ['./one-variant-only-notification.component.scss'],
+})
+export class OneVariantOnlyNotificationComponent {
```
kyubisation, line 10: I'm aware I'm bikeshedding, but maybe rename this to VariantLimitationComponent? What is your opinion on this?

src/showcase-merge/app/shared/component-viewer/one-variant-only-notification/one-variant-only-notification.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  @Input()
+  showcaseMetaEntry: ShowcaseMetaEntry;
+
+  get variantOnly(): 'standard' | 'lean' {
+    return this.showcaseMetaEntry.variantOnly;
+  }
+
+  oppositeVariant(variantOnly: 'standard' | 'lean'): 'standard' | 'lean' {
```
kyubisation, line 18: Why not use a getter for this?

src/showcase-merge/app/shared/library-viewer/library-viewer.component.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+import { ShowcaseMetaLibrary } from '../meta';
+
+@Component({
+  selector: 'sbb-library-viewer',
+  templateUrl: './library-viewer.component.html',
+})
+export class LibraryViewerComponent {
```
kyubisation, line 12: Related to my library vs package comment; Maybe PackageViewerComponent?

src/showcase-merge/app/shared/meta.ts
```diff
@@ -0,0 +1,155 @@
...
+        },
+      ],
+    },
+  ];
+
+  static libraries(): ShowcaseMetaLibrary[] {
+    return this.showcaseMetaLibraries;
+  }
+
+  static findByLibraryName(libraryName: string): ShowcaseMetaLibrary {
```
kyubisation, line 140: Since we only have static data, should we maybe use maps instead of arrays in order to avoid these lookups?

kyubisation (changes requested):
See comments. Maybe we should discuss this together.

src/showcase-merge/app/shared/component-viewer/one-variant-only-notification/one-variant-only-notification.component.html
jeripeierSBB: textexpand component is not usable inside a notification because a color is defined and would have to be overwritten and "show more" for me is not the best term . Tooltip is not the best because there is already an info symbol and it looks somehow bad. Other option is to create own own textexpand like component or as a small improvement, make info text 13px

kyubisation (approved):
LGTM 👍

PR #772, opened 2021-03-24
fix: update xlf files to fix interpolated entries

kyubisation:
The Angular i18n extractor changed the output format for interpolation
elements, which causes consumers to break when using our xlf files.


jeripeierSBB (approved):
LGTM

PR #775, opened 2021-03-26
build: switch browser provider

kyubisation:
We previously used the browsers from the bazelbuild/rules_webtesting library.
Since this package has not been updated in a while, the browsers have become
outdated.
This PR switches the browser provider to angular/dev-infra-private-builds, which should
be acceptable, according to angular/components#22335


jeripeierSBB (approved):
LGTM

PR #789, opened 2021-04-14
fix(business): update SbbTableDataSource data on disconnected state

kyubisation:


src/angular-business/table/table/table-data-source.ts
```diff
@@ -372,6 +391,10 @@ export class SbbTableDataSource<
    * @docs-private
    */
   connect() {
+    if (!this._renderChangesSubscription) {
```
jeripeierSBB, line 394: Don't we have to support disconnect?

  /**
   * Used by the SbbTable. Called when it connects to the data source.
   * @docs-private
   */
  connect() {
    if (!this._renderChangesSubscription) {
      this._updateChangeSubscription();
    }

    return this._renderData;
  }

src/angular-business/table/table/table-data-source.ts
```diff
@@ -372,14 +391,21 @@ export class SbbTableDataSource<
...
+      this._updateChangeSubscription();
+    }
+
     return this._renderData;
   }
 
   /**
    * Used by the Table. Called when it is destroyed. No-op.
```
jeripeierSBB, line 402:
```suggestion
   * Used by the SbbTable. Called when it disconnects from the data source.
```

src/angular-business/table/table/table-data-source.ts
kyubisation: You're absolutey right

jeripeierSBB (approved):
LGTM

But I did no test :)

PR #794, opened 2021-04-22
fix(business): hide sidebar scrollbar on iOS devices

jeripeierSBB:
Sets scrollbar on iOS devices invisible which solves a bug where a scrollbar
is wrongly visible for a short time (until clicking or scrolling) inside the sidebar.

### DIFF
--- a/src/angular-core/styles/typography/_scrollbar.scss
+++ b/src/angular-core/styles/typography/_scrollbar.scss
@@ -27,7 +27,7 @@ $trackColor: $sbbColorMilk;
     }
 
     @supports (-webkit-touch-callout: none) {
-      // Only visible on iOS devices.
+      // Only applied on iOS devices.
       // Sets scrollbar on iOS devices invisible which solves a bug where a scrollbar
       // is wrongly visible for a short time (until clicking or scrolling) inside the sidebar.
       &::-webkit-scrollbar {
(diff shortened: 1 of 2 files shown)


src/angular-core/styles/typography/_scrollbar.scss
```diff
@@ -26,6 +26,16 @@ $trackColor: $sbbColorMilk;
       height: 0.5rem;
     }
 
+    @supports (-webkit-touch-callout: none) {
+      // Only visible on iOS devices.
```
kyubisation, line 30: This seems to be an odd phrasing for something hidden on iOS. See my suggestion or correct me, if I mistunderstood it.
```suggestion
      // Only applied on iOS devices.
```

src/angular/styles/_typography.scss
```diff
@@ -1936,6 +1936,16 @@ $trackColor: $sbbColorMilk;
     height: 0.5rem;
   }
 
+  @supports (-webkit-touch-callout: none) {
+    // Only visible on iOS devices.
```
kyubisation, line 1940:
```suggestion
    // Only applied on iOS devices.
```

kyubisation (commented):
Thank you for the fix 👍 
See my comment.

PR #799, opened 2021-05-05
fix(business): prevent submit event in chips when pressing enter with content

kyubisation:
Currently if enter is pressed in the chips input, the submit event from the surrounding
form is triggered. In order to prevent this from always happening, we check if content
has been entered in the input and if yes, we prevent the form submit. If the inner
input has no value, then the enter press will trigger the form submit.


jeripeierSBB (approved):
LGTM

PR #803, opened 2021-05-12
fix(business): avoid metadata inlining for the header

### LINKED ISSUE #798: Bug: business stackblitz examples throw error
https://stackblitz.com/angular/nndgexjynej?file=src%2Fapp%2Fforms-chip-input-example.ts

throws Error

Error in /turbo_modules/@sbb-esta/angular-business@11.1.3/bundles/angular-business-header.umd.js (1018:17)
SbbHeader is not defined

kyubisation:
Due to SbbHeader being imported as type and directly used as a parameter
type in SbbHeaderMenu, the decorater metadata contained a reference to
SbbHeader, which was not actually available.

Closes #798


jeripeierSBB (approved):
LGTM

PR #819, opened 2021-05-31
refactor(maps): replacing the esri-loader with @arcgis/core ES-module

kyubisation:
Since ArcGIS API for Javascript Version 4.18 it's easier to work with
Esri-Types: https://developers.arcgis.com/javascript/latest/es-modules/.
This update is replacing the esri-loader module with the new
@arcgis/core ES-module.

BREAKING CHANGE: esri-loader is replaced with @arcgis/core as a
peer dependency for @sbb-esta/angular-maps and usage has been
updated accordingly.


kyubisation:
@mstankala, @ajoos: We have switched to esbuild for building the showcase(s), which has reduced the build time from minutes to a few seconds. The bundle size still has increased tremendously, but it should be fine for now.

https://angular-staging.app.sbb.ch/819/maps/introduction/getting-started

jeripeierSBB (commented):
Bildschirmfoto 2021-06-01 um 09 08 58

@mstankala Do you know this warning and maybe how to fix it?

ajoos:
> 
> 
> @mstankala, @ajoos: We have switched to esbuild for building the showcase(s), which has reduced the build time from minutes to a few seconds. The bundle size still has increased tremendously, but it should be fine for now.
> 
> https://angular-staging.app.sbb.ch/819/maps/introduction/getting-started

That's great news, glad to hear this change also improves the showcase build as a whole. What would be an approach to decrease the bundle size?

ajoos:
> 
> 
> Bildschirmfoto 2021-06-01 um 09 08 58
> 
> @mstankala Do you know this warning and maybe how to fix it?

Here's a link to a discussion of the problem: https://community.esri.com/t5/arcgis-api-for-javascript/warning-version-mismatch-detected-between-arcgis-api-for/m-p/1054789 - I believe the workaround provided there doesn't work for us. @mstankala have you experienced this before (with DiDok for example)?

kyubisation:
Thank you for the feedback 👍 
I'm afraid we can't really decrease total bundle size, unless esri removes their dependency on moment.js. However I think that still won't massively reduce the bundle size, as @arcgis/core and the features we use from it are simply large parts/bundles.

jeripeierSBB (approved):
LGTM

PR #820, opened 2021-06-02
fix: fix import errors if using primary imports

jeripeierSBB:
Starting with release 12.0.0, importing from `@sbb-esta/angular-business`,
 `@sbb-esta/angular-public`and `@sbb-esta/angular-core` are producing
 compile errors. This fix re-allows primary imports.

We generally recommend to use secondary module imports
 like `@sbb-esta/angular-business/radio-button`,
 which was not affected by this issue.


kyubisation (approved):
LGTM 👍

PR #839, opened 2021-07-07
fix: allow custom tabindex on expansion panel header

### LINKED ISSUE #838: Bug: Expansion Panel Header configurable tabindex
**Describe the bug**
When the expansion panel header is not disabled, it has always `tabindex=0`. This hinders a natural tabindex flow for long forms that are structured using accordions - e.g.:
image

It should be possible to set `<sbb-expansion-panel-header tabindex="-1">` (or similar) to exclude the expansion panel header from the tab index flow.

kyubisation:
Closes #838


jeripeierSBB (approved):
LGTM

PR #847, opened 2021-07-19
refactor: enable `noImplicitOverride` and remove mixin types

kyubisation:


jeripeierSBB (commented):
I suggest to remove exports of *Ctor types in src/angular because we don't need any deprecation here.

jeripeierSBB (approved):
LGTM

PR #848, opened 2021-07-19
refactor: remove unused imports

kyubisation:


tslint.json
```diff
@@ -151,6 +151,8 @@
     "no-switch-case-fall-through": true,
     "no-unnecessary-initializer": true,
     "no-unused-expression": true,
+    // Uncomment to following line and run `yarn tslint --fix` to remove unused imports
+    // "no-unused-variable": [true, {"ignore-pattern": "^_"}],
```
jeripeierSBB, line 155: Why don't we activate it as a standard?

kyubisation: Because we have some unused constructor parameter that we can't remove yet and the _ does not seem to work. But I can check it quickly.

jeripeierSBB (approved):
LGTM except for the build errors

PR #849, opened 2021-07-21
perf: aggregate refactor migration

kyubisation:
In order to increase performance of our many refactor migrations, this
PR aims to aggregate the refactor migrations. This ensures the template
DOM tree walking is reduced.


jeripeierSBB (approved):
LGTM

PR #851, opened 2021-07-22
build: add schematics for selector groups for metrics

kyubisation:


jeripeierSBB (approved):
LGTM, maybe wait with merge until the metrics tool works with

PR #853, opened 2021-07-26
feat: add css class for normal error flow inside sbb-form-field

kyubisation:
We are adding the css class `sbb-form-field-flexible-errors`, which can be placed on `sbb-form-field` or any ancestor, which will remove the reserved space for errors and display errors in the normal context flow.
This also deprecates the css class `sbb-form-field-errorless`, which will be removed in a future version.


jeripeierSBB (approved):
LGTM, nice work and I think a good update to our library

PR #859, opened 2021-08-12
refactor(angular): consider options length when setting trigger css classes

jeripeierSBB:
The trigger css classes `sbb-input-with-open-panel` and `sbb-input-with-open-panel-above` were set only depending on open state of the autocomplete panel. 
If there are no options inside the autocomplete, the autocomplete is not shown to the user, but previously the border-radius was already set like it is open. With this change, border-radius is set at the right time.


src/angular/autocomplete/autocomplete.ts
```diff
@@ -111,11 +111,11 @@ export class SbbAutocomplete extends _SbbAutocompleteBase implements AfterConten
   _keyManager: ActiveDescendantKeyManager<SbbOption>;
 
   /** Whether the autocomplete panel should be visible, depending on option length. */
-  showPanel: boolean = false;
+  showPanel: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
```
kyubisation, line 114: I'd like to avoid changing the API surface.
Would it be an option to have an additional property (e.g. _showPanel) for the subject and convert showPanel into a getter?

jeripeierSBB: good point, changed it

kyubisation (approved):
LGTM 👍

PR #867, opened 2021-08-26
fix: don't reset height for sbb-textarea

### LINKED ISSUE #862: Bug: Textarea jumps/leads to scrolling on input
**Describe the bug**
when having longer input (many lines) in the textarea it sometimes leads to unintended scrolling.

**To Reproduce**
1. Go to https://angular.app.sbb.ch/public/components/textarea
2. Click on "Showcase"
3. Enter multiple lines
4. Scroll so that only last few lines are visible
5. Enter empty lines and text
6. See that sometimes the scrollbar jumps (not easy to reproduce)

**Expected behavior**
Scrolling should be consistent and not jumping up and down

**Screenshots**
jumping-textarea

 - OS: Win10
 - Browser Chrome  92.0.4515.131 (Offizieller Build) (64-Bit)
 - Version SACL 12.3

**Additional context**
In our project (with perfect-scrollbar) this can always/easily be reproduced!
Maybe related: https://stackoverflow.com/questions/56329625/preventing-textarea-scroll-behaviour-in-chrome-after-newline-added

kyubisation:
Closes #862

### DIFF
--- a/src/angular-public/typography.scss
+++ b/src/angular-public/typography.scss
@@ -15,6 +15,7 @@
 @include sbbTable();
 @include cdk.a11y-visually-hidden();
 @include cdk.overlay();
-@include cdk.text-field();
+@include cdk.text-field-autosize();
+@include cdk.text-field-autofill();
 @include sbbOverlay();
 @include sbbScrollbar();
--- a/src/angular-public/typography.scss
+++ b/src/angular-public/typography.scss
@@ -15,5 +15,6 @@
 @include sbbTable();
 @include cdk.a11y-visually-hidden();
 @include cdk.overlay();
+@include cdk.text-field();
 @include sbbOverlay();
 @include sbbScrollbar();
--- a/src/showcase/styles.scss
+++ b/src/showcase/styles.scss
@@ -13,7 +13,8 @@ body {
 @include sbbLink();
 @include cdk.a11y-visually-hidden();
 @include cdk.overlay();
-@include cdk.text-field();
+@include cdk.text-field-autosize();
+@include cdk.text-field-autofill();
 @include sbbOverlay();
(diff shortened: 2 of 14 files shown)


jeripeierSBB:
I think in the merge showcase, the fix does not resolve the problem?
 
https://angular-staging.app.sbb.ch/867-merge/angular/components/textarea/examples

src/showcase/styles.scss
```diff
@@ -13,6 +13,7 @@ body {
 @include sbbLink();
 @include cdk.a11y-visually-hidden();
 @include cdk.overlay();
+@include cdk.text-field();
```
jeripeierSBB, line 16: import seems to be deprecated:

```
// @deprecated Use 'autosize' and 'autofill' instead.
@mixin text-field {
  @include text-field-autosize();
  @include text-field-autofill();
}
```

src/angular-public/typography.scss
```diff
@@ -15,5 +15,6 @@
 @include sbbTable();
 @include cdk.a11y-visually-hidden();
 @include cdk.overlay();
+@include cdk.text-field();
```
jeripeierSBB, line 18: import seems to be deprecated:

```
// @deprecated Use 'autosize' and 'autofill' instead.
@mixin text-field {
  @include text-field-autosize();
  @include text-field-autofill();
}
```

jeripeierSBB (changes requested):
Maybe we should also enhance the markdown file?

Futhermore I think we should change the PR title to a refactor, because it doesn't really solve the issue, but it's better code

kyubisation:
I updated the markdown files.
I would disagree; Previously the necessary CDK class was missing from the typography, which was causing an issue for the autosizing. This PR fixes that problem.

PR #870, opened 2021-08-27
fix: display header menu trigger in appropriate circumstances

kyubisation:


jeripeierSBB (approved):
LGTM, I'm just asking myself whether it would be better to add a test for this case

PR #874, opened 2021-09-06
refactor: tooltip

### LINKED ISSUE #790: Bug: Tooltip delay don't work if injected by SBB_TOOLTIP_DEFAULT_OPTIONS
**Describe the bug**
The tooltip delays don't work as specified if injected by SBB_TOOLTIP_DEFAULT_OPTIONS.
On the other hand if the delays are set by attribute it works as expected.

**To Reproduce**
See Reproduction: https://stackblitz.com/edit/angular-uzgvep-1bdl3o

kyubisation:
Closes #790


src/components-examples/angular/tooltip/tooltip-hover/tooltip-hover-example.html
```diff
@@ -0,0 +1,20 @@
+<h4>Tooltip Component Usage</h4>
+<sbb-tooltip trigger="hover" [hoverShowDelay]="hoverShowDelay" [hoverHideDelay]="hoverHideDelay">
+  <span>This tooltip was shown on hover.</span>
+</sbb-tooltip>
+
+<h4>Tooltip Directive Usage</h4>
+<span
```
jeripeierSBB, line 7: better use `<p>` here because it would scale in 4k and 5k (span does not and stays very small).

src/components-examples/angular/tooltip/tooltip-simple/tooltip-simple-example.html
```diff
@@ -0,0 +1,16 @@
+<sbb-tooltip>{{ tooltipContent }}</sbb-tooltip>
+<h4>Properties</h4>
+<sbb-form-field label="Tooltip text content">
+  <input type="text" [(ngModel)]="tooltipContent" sbbInput />
+</sbb-form-field>
+<div>
```
jeripeierSBB, line 6: I'd suggest to use `<p>` here so the 4k and 5k showcase looks better (scaled font size)

jeripeierSBB (changes requested):
Some more findings:
- it's not possible to reach the close icon or every other tooltip content by keyboard navigation
- merge symbols are missing (Were there really no component renames which have to be tracked by automatic migration?)
- In the showcase merge there are still usages of angular-business tooltip
- In components examples there is a angular-public tooltip usage
- toggling with space bar or enter key does not work (before merge it did)
- space between tooltip arrow and text in directive usage is too small, especially in standard. in 5k it does even collide

src/angular/tooltip/tooltip.ts
```diff
@@ -0,0 +1,884 @@
...
+  animations: [sbbTooltipAnimations.tooltipState],
+  host: {
+    class: 'sbb-tooltip-component',
+    // Forces the element to have a layout in IE and Edge. This fixes issues where the element
+    // won't be rendered if the animations are disabled or there is no web animations polyfill.
+    '[style.zoom]': '_visibility === "visible" ? 1 : null',
+    'aria-hidden': 'true',
+  },
+})
+export class TooltipComponent extends _TooltipComponentBase {
```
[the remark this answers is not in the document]
kyubisation: I would like to avoid that, since we previously had a SbbTooltipComponent. Also this is not a component that should be used.

src/angular/schematics/ng-add/migrations/merge-refactor/tooltip-migration.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  protected _migrate(element: MigrationElement) {
+    const icon = element.findProperty('icon');
+    if (icon) {
+      icon.remove();
+      this._tooltipMigrationFailedPartially = true;
+      element.insertStart(
+        `<!-- TODO: Unable to determine custom icon from icon "${icon.attribute.value}". ` +
+          'Please manually select a custom indicatorIcon: https://angular.app.sbb.ch/angular/components/button -->'
```
jeripeierSBB, line 37:
```suggestion
          'Please manually select a custom indicatorIcon: https://angular.app.sbb.ch/angular/components/tooltip -->'
```

src/angular/tooltip/tooltip-wrapper.scss
```diff
@@ -16,31 +21,31 @@
...
 }
 
 .sbb-tooltip-close-icon {
-  border: pxToEm(1) solid $sbbColorGrey;
+  border: pxToEm(1) solid $sbbColorStorm;
   border-radius: 50%;
+  color: $sbbColorGranite;
   float: right;
   position: absolute;
   // padding + margin-top + border
```
jeripeierSBB, line 40:
```suggestion
```

src/angular/tooltip/tooltip-wrapper.scss
```diff
@@ -51,7 +56,11 @@
 
   .sbb-tooltip-panel-above & {
     // padding + border
```
jeripeierSBB, line 58:
```suggestion
```

jeripeierSBB (changes requested):
- "esc" key seems to work not anymore
- hover color and focus color of cross icon in lean is not set

jeripeierSBB (approved):
LGTM

If you like to be outstandig ;), a unit test for the "esc" case would be very nice

jeripeierSBB:
two more findings: 
- tooltip "Tooltip Custom Content" example does focus the link on open instead of the cross icon. toggling is not possible.
- the focus style of the cross icon does somehow kick in with a delay.

kyubisation:
The logic focuses the first focusable element. I think that is the correct behavior for a persistent overlay.
The delay is probably caused by waiting for the animation to finish. This should ensure that the content is actually ready.

PR #877, opened 2021-09-07
fix: ensure correct font-family

kyubisation:
arcgis defines its own font-family for :root (html). In order to reliably overwrite any other definition, we have added !important to the font-family.


jeripeierSBB (commented):
There is also a second value which is defined by arcgis: 
`text-rendering: optimizeLegibility`

Should we overwrite that too?
https://developer.mozilla.org/de/docs/Web/CSS/text-rendering

kyubisation:
Good question. As it doesn't seem to affect our font/text, I would ignore it. I would like to avoid having to overwrite all rules.

PR #883, opened 2021-09-23
fix(business): vertically center chip input label text

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
However, the merged variant seems off too. The height should be 36, but it is only 34 for some reason.

PR #903, opened 2021-11-02
refactor(angular): configure typography for ng-add command

jeripeierSBB:


src/angular/schematics/ng-add/setup-project.ts
```diff
@@ -187,3 +182,91 @@ function validateDefaultTargetBuilder(
...
+  return async (tree: Tree, context: SchematicContext) => {
+    const workspace = await getWorkspace(tree);
+    const project = getProjectFromWorkspace(workspace, options.project);
+    const targetOptions = getProjectTargetOptions(project, 'build');
+    const shouldBeLeanVariant = options.variant === 'lean (previously known as business)';
+
+    if (!targetOptions?.index) {
+      if (shouldBeLeanVariant) {
+        context.logger.error(
+          `Could not find index.html to configure design variant. If you like to use the lean design variant, please add 'sbb-lean' attribute to the <html> tag.`
```
kyubisation, line 219: This should be a class: `class="sbb-lean"`

kyubisation (changes requested):
LGTM 👍 
However, `sbb-lean` is used as a class, not as an attribute.

kyubisation (approved):
LGTM 👍

PR #904, opened 2021-11-02
refactor(angular): replace i18n tags by $localize

jeripeierSBB:


src/angular/alert/alert.html
```diff
@@ -16,10 +16,6 @@
     (click)="_handleDismiss($event)"
   >
     <sbb-icon svgIcon="kom:cross-small"></sbb-icon>
-    <span
-      i18n="Hidden button label to close the alert@@sbbAlertCloseAlert"
-      class="cdk-visually-hidden"
-      >Close message</span
-    >
+    <span class="cdk-visually-hidden">{{_labelClose}}</span>
```
kyubisation, line 19: Doesn't prettier format this as `{{ _labelClose }}`?

kyubisation (approved):
LGTM 👍 
I'm a bit confused about the formatting in the templates (`{{example}}` instead of `{{ example }}`)

src/angular/alert/alert.html
jeripeierSBB: it seems that prettier ignores this (I can write whatever i want and nothing changes). But I fixed it manually

PR #913, opened 2021-11-08
refactor: fix ellipsis for buttons

kyubisation:


src/angular/styles/_typography.scss
```diff
@@ -997,7 +997,7 @@ fieldset,
   max-width: none;
   min-width: 0;
   text-decoration: none;
-  display: inline-block;
+  display: inline-flex;
```
kyubisation, line 1000:
```suggestion
  display: inline-block;
```

jeripeierSBB (commented):
If the width is small (but still above min-width), the arrow is displayed inside the text.

on general question: why is there a max-width for the button?

PR #928, opened 2021-11-17
refactor: optimize form field styles

kyubisation:


jeripeierSBB (approved):
LGTM

Do you remember why we used !important? Maybe this could be removed now
Bildschirmfoto 2021-11-24 um 08 53 17

PR #931, opened 2021-11-19
refactor: optimize checkbox/radio button (panel) styles

kyubisation:


jeripeierSBB (changes requested):
Generally looks good (I admit that I have not verified every single property), but 5k design seems to be broken (testes with checkbox)

jeripeierSBB (approved):
LGTM

PR #936, opened 2021-11-25
fix(showcase): fix stackblitz i18n and RouterModule

jeripeierSBB:


src/showcase-merge/assets/stack-blitz/src/main.ts
```diff
@@ -4,6 +4,12 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
 import {AppModule} from './app/app.module';
 import {environment} from './environments/environment';
 
+import '@angular/localize/init';
```
kyubisation, line 7: This should be before the other imports

kyubisation (approved):
LGTM 👍

PR #947, opened 2021-11-30
refactor: optimize tooltip styles

kyubisation:


src/angular/tooltip/tooltip.scss
```diff
@@ -45,43 +40,40 @@ $tooltipShadowArrowSizeLean: 13.333;
...
 .sbb-tooltip-shadow {
-  height: pxToEm(8);
-  background-color: rgba(0, 0, 0, 0.15);
-  border-radius: 0 0 pxToEm(2) pxToEm(2);
+  height: var(--sbb-box-shadow-width);
+  background-color: var(--sbb-box-shadow-color);
+  border-radius: 0 0 var(--sbb-border-radius) var(--sbb-border-radius);
+
+  :where(html.sbb-lean) & {
+    border-radius: 0 0 pxToRem(1) pxToRem(1);
```
jeripeierSBB, line 75: I think 2px would be fine here too.

General problem, the shadow should be taller:
height: calc(var(--sbb-box-shadow-width) + var(--sbb-border-radius));
margin-top: calc(-1 * var(--sbb-border-radius);

and as a following problem, stacking of arrow, content and shadow has to be resolved, probably by z-index and position absolute.

jeripeierSBB (changes requested):

src/angular/tooltip/tooltip.scss
jeripeierSBB: Bildschirmfoto 2021-12-01 um 09 15 36

src/angular/tooltip/tooltip.scss
```diff
@@ -15,106 +10,102 @@ $tooltipShadowArrowSizeLean: 13.333;
   font-size: var(--sbb-font-size);
   line-height: var(--sbb-line-height);
 
-  &::after {
-    top: 0;
+  &::after,
+  &::before {
+    // Slightly move down arrow to avoid artifacts on certain zoom levels
+    top: calc(0.25 * var(--sbb-border-width));
```
jeripeierSBB, line 16:
```suggestion
    top: calc(0.5 * var(--sbb-border-width));
```

jeripeierSBB (approved):
LGTM

PR #960, opened 2021-12-03
refactor: optimize processflow styles

kyubisation:


src/angular/processflow/step-header.scss
```diff
@@ -46,7 +40,7 @@
```
jeripeierSBB, line 40: sbb-step-text-label: a case for @include ellipsis() ?

src/angular/processflow/step-header.scss
```diff
@@ -77,11 +67,9 @@
...
-  &::after {
-    left: 0;
-    width: 100%;
-  }
+.sbb-step-label-selected::after {
+  left: 0;
+  width: 100%;
 }
 
 .sbb-step-label-active.sbb-step-label-locked {
```
jeripeierSBB, line 75: .sbb-step-header:hover & {
    color: var(--sbb-color-text);
  }
it looks that in locked step the hover color has to be overwritten, therefore it should be color-text. or not?

jeripeierSBB (changes requested):
it's somehow possible to display hover color on a locked step, weird.
Bildschirmfoto 2021-12-06 um 11 41 07

jeripeierSBB (approved):
LGTM

PR #962, opened 2021-12-03
refactor: optimize lightbox/dialog styles

kyubisation:


src/angular/lightbox/lightbox.scss
```diff
@@ -38,61 +38,46 @@
...
-
   // Remove the top margin from headings if it is the first element inside the content.
-  h1,
-  h2,
-  h3,
-  h4 {
-    &:first-child {
-      margin-top: 0;
-    }
+  :is(h1, h2, h3, h4):first-child {
```
jeripeierSBB, line 47:
```suggestion
  :is(h1, .sbb-headline1, h2, .sbb-headline2, h3, .sbb-headline3, h4, .sbb-headline4, p, .sbb-copy, .sbb-text-lead):first-child {
```

jeripeierSBB: like it is in expansion panel. maybe we can add the selectors to a sass var?

src/angular/lightbox/lightbox.scss
```diff
@@ -38,61 +38,46 @@
...
+    margin-top: 0;
   }
 }
 
 .sbb-lightbox-title {
-  padding: pxToRem(21) pxToRem(44) pxToRem(17) pxToRem(20);
+  // The right padding considers original padding, plus additional padding
```
jeripeierSBB, line 53: Can we not express that with calc and icon width?

jeripeierSBB (changes requested):
idea: use <h3></h3> (or other heading) in example for sbbDialogTitle to achieve a better look of dialog in standard variant

jeripeierSBB (approved):
LGTM

PR #964, opened 2021-12-03
refactor: optimize datepicker styles

kyubisation:
@jeripeierSBB I would like to discuss the size of the "spickel" in relation to the tooltip one.

### DIFF
--- a/src/angular/datepicker/datepicker-content/datepicker-content.scss
+++ b/src/angular/datepicker/datepicker-content/datepicker-content.scss
@@ -1,24 +1,6 @@
 @import '../../styles/common';
 
-$datePickerContentOffset: 17;
-$datePickerToggleArrowColor: var(--sbb-color-grey);
 $datePickerCalendarBorderWidth: 2;
-$datePickerCalendarBg: var(--sbb-color-white);
-$datePickerCalendarHeaderColor: var(--sbb-color-metal);
-$datepickerCalendarLineColor: var(--sbb-color-platinum);
-$datePickerCalendarHeaderHeight: 68;
-$datePickerCalendarTableContentPaddingWidthMobile: 11;
-$datePickerCalendarTableContentPaddingWidth: 18;
-$datePickerCalendarTableActiveColor: var(--sbb-color-red125);
-$datePickerCalendarTableRangeBgColor: var(--sbb-color-milk);
-
-// Lean variables
-$datePickerContentOffsetLean: 8;
-$datepickerCalendarLineColorLean: var(--sbb-color-cloud);
-$datePickerCalendarHeaderHeightLean: 36;
-$datePickerCalendarHeaderColorLean: var(--sbb-color-iron);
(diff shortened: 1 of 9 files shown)


src/angular/styles/_typography.scss
```diff
@@ -288,6 +329,28 @@
...
+    --sbb-datepicker-calender-header-color: var(--sbb-color-iron);
+    --sbb-datepicker-calender-line-color: var(--sbb-color-cloud);
+    --sbb-datepicker-calender-table-content-padding-width: #{pxToRem(8)};
+    // TODO: Discuss size with jeripeierSBB
+    --sbb-datepicker-calender-arrow-width: #{pxToRem(10)};
+    --sbb-datepicker-calender-width: #{pxToRem(303)};
+    --sbb-datepicker-calender-height: #{pxToRem(303)};
+    --sbb-datepicker-calender-controls-padding: #{pxToRem(8)};
+    --sbb-datepicker-calender-body-cell-padding: #{pxToRem(16.5)};
+    --sbb-datepicker-calender-body-cell-disabled-offset: #{pxToRem(17)};
```
jeripeierSBB, line 351:
```suggestion
```
should be 12 as in standard mobile definition, so therefore this definition should not be needed

src/angular/datepicker/datepicker-content/datepicker-content.scss
```diff
@@ -1,24 +1,6 @@
 @import '../../styles/common';
 
-$datePickerContentOffset: 17;
-$datePickerToggleArrowColor: var(--sbb-color-grey);
 $datePickerCalendarBorderWidth: 2;
```
jeripeierSBB, line 3:
```suggestion
```

src/angular/styles/_typography.scss
```diff
@@ -216,6 +246,7 @@
 
   // Shadow
   --sbb-box-shadow-width: calc(var(--sbb-border-width) * 4);
+  --sbb-box-shadow-after-width: calc(var(--sbb-border-width) * 2);
```
jeripeierSBB, line 249: This should no be a var in global scope. In the datepicker usage it can be any size, the real width is determined by clip path. We can als use --sbb-box-shadow-width or just no var ;)

jeripeierSBB (changes requested):
- lean: left offset of arrow wrong
- lean: disabled calendar slash at wrong position

color grey (standard) and iron (lean) are often a pair. I think here we can maybe add a var (global or only datepicker scope)

the last two examples have a form input label called "Datepicker", but there is no datepicker.

jeripeierSBB (changes requested):
- why --sbb-box-shadow-arrow-width? I think we discussed to switch it just to --sbb-box-shadow-width?
- Datepicker label in two examples wrong
- Renaming of calender to calendar
- Safari arrow shadow doesn't look good in standard design. We have probably to accept this or switch to tooltip implementation

PR #972, opened 2021-12-08
fix: avoid closing tooltip on clipping parent container

kyubisation:
Previously the tooltip would close when it clipped the scrollable
context/parent. This would lead to the tooltip immediately closing on
scrolling when above such a border. This PR changes this behavior to
only close the tooltip when the origin of the tooltip is scrolled outside
the view.


jeripeierSBB (approved):
LGTM

PR #977, opened 2021-12-09
refactor: weak accent color

jeripeierSBB:


src/angular/styles/_typography.scss
```diff
@@ -1687,6 +1674,10 @@ textarea {
   user-select: none;
   position: relative;
 
+  :where(html.sbb-lean) & {
```
kyubisation, line 1677: Is this necessary? grey should be granite in this context, right?

kyubisation (commented):
LGTM 👍 
One question about a lean usage.

kyubisation (approved):
LGTM 👍

PR #994, opened 2021-12-16
fix: reduce space of buttons inside table cells

jeripeierSBB:


src/angular/styles/_typography.scss
```diff
@@ -1857,6 +1872,34 @@ textarea {
     }
   }
 
+  // Position buttons and contextmenu over padding of cell
+  // Only in Lean: Additionally reduce size of buttons to fit into cell
```
jeripeierSBB, line 1876:
```suggestion
  // Only in lean design: Additionally reduce size of buttons to fit into cell
```

PR #1001, opened 2021-12-23
fix(angular/datepicker): reset model of connected datepicker if datepicker value changes

### LINKED ISSUE #993: Bug: Datepicker does not reset second picker in connected mode correctly
**Describe the bug**
When entering a date in the first datepicker that is after the date that was entered in the second datepicker, the second datepicker visually removes the date, but it stays in the model. 

connected-datepicker

**To Reproduce**
See animation above with the showcase.

**Expected behavior**
Either the model should be reset as well (preferred) or the old value should be retained visually.

 - OS: Win10
 - Browser: Chrome
 - Version: SACL v12 and v13.next (!)

jeripeierSBB:
Closes #993


kyubisation (approved):
LGTM 👍

PR #1006, opened 2022-01-07
feat(angular/header): add environment `local` with color `autumn`

### LINKED ISSUE #893: Add local to Header Environment
Add `.sbb-header-environment-local` like `.sbb-header-environment-dev`, ...

We use and suggest gray as color for local environments.

Having local seems to be a common case.

kyubisation:
Closes #893


jeripeierSBB (approved):
LGTM

PR #1011, opened 2022-01-11
fix: improve accessibility of datepicker and toggle

mhaertwig:
- allow defining aria-labels for nextDay and prevDay buttons in datepicker
- use aria-checked for role "radio" in toggle, since aria-expanded is not allowed for this role. (see https://www.w3.org/TR/wai-aria-1.1/#radio)
- add aria-label for chips drag and drop example


src/angular/datepicker/datepicker/datepicker.ts
```diff
@@ -212,6 +212,12 @@ export class SbbDatepicker<D> implements OnDestroy {
     );
   }
 
+  /** The next day button's aria-label */
+  @Input() nextDayAriaLabel: string;
```
kyubisation, line 216: Add default values with the following pattern:
```ts
typeof $localize === 'function'
      ? $localize`:Description@@sbbDatepicker...:...`
      : '...'
```

src/components-examples/angular/datepicker/datepicker-simple-reactive/datepicker-simple-reactive-example.html
```diff
@@ -1,6 +1,8 @@
 <sbb-form-field label="Datepicker">
   <sbb-datepicker
     [arrows]="arrows.value"
+    nextDayAriaLabel="Select next date"
```
kyubisation, line 4: Maybe move these examples to the date range examples

kyubisation (changes requested):
Thanks for the contribution 👍 
As discussed

mhaertwig:
Split into #1014 and #1015.

PR #1013, opened 2022-01-12
fix(angular/toggle): use aria-checked for role "radio" in toggle

mhaertwig:
aria-expanded is not allowed for this role.
see https://www.w3.org/TR/wai-aria-1.1/#radio


mhaertwig:
Sorry, but I have to correct my previous statement. [MDN](https://developer.mozilla.org/en-US/docs/web/Accessibility/ARIA/Attributes/aria-checked) says the following:

> Note: Where possible use an HTML <input> element with type="checkbox" as this element has built in semantics and does not require ARIA attributes.

According to [WAI](https://www.w3.org/TR/wai-aria-1.1/#radio) `aria-expanded` is not valid for the radio role since screen readers may not understand it. However, we also don't need an `aria-checked` attribute. So we could either
1. leave `aria-expanded` and hope some screen readers understand what we mean
2. or remove the `aria-expanded` (and `aria-checked`) attributes completely.

What do you think?

mhaertwig:
I've investigated this a bit deeper and suggest going with option 2 (removing `aria-checked` and `aria-expanded`).
- `aria-expanded` is not allowed on radio elements and thus not understood by screen readers.
- `aria-checked` is not necessary on native elements (only on custom widgets without an underlying native element).
- Changing the role, e.g. to "tab" is not suitable since the options onside one toggle group can have associated input or not.
- The current usage of `aria-controls` is the right way to go.

jeripeierSBB (approved):
LGTM, thanks

PR #1022, opened 2022-01-14
refactor(angular): remove coercion members

kyubisation:
Replaces all usages of the `ngAcceptInputType` members with type declarations directly on the setters.

### DIFF
--- a/src/angular/datepicker/calendar-body/calendar-body.ts
+++ b/src/angular/datepicker/calendar-body/calendar-body.ts
@@ -1,4 +1,3 @@
-import { BooleanInput } from '@angular/cdk/coercion';
 import {
   ChangeDetectionStrategy,
   Component,
@@ -110,6 +109,4 @@ export class SbbCalendarBody {
         });
     });
   }
-
-  static ngAcceptInputType_allowDisabledSelection: BooleanInput;
 }
--- a/src/angular/file-selector/file-selector.ts
+++ b/src/angular/file-selector/file-selector.ts
@@ -91,7 +91,7 @@ export class SbbFileSelector
   get multiple(): boolean {
     return this._multiple;
   }
-  set multiple(value: boolean) {
+  set multiple(value: BooleanInput) {
     this._multiple = coerceBooleanProperty(value);
   }
   private _multiple = false;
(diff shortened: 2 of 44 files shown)


src/angular/datepicker/calendar-body/calendar-body.ts
```diff
@@ -110,6 +109,4 @@ export class SbbCalendarBody {
         });
     });
   }
-
-  static ngAcceptInputType_allowDisabledSelection: BooleanInput;
```
jeripeierSBB, line 114: Are there really no change required here:
```
/** Whether to allow selection of disabled cells. */
  @Input() allowDisabledSelection: boolean = false;
```

src/angular/file-selector/file-selector.ts
```diff
@@ -232,9 +232,4 @@ export class SbbFileSelector
       file1.lastModified === file2.lastModified
     );
   }
-
-  static ngAcceptInputType_capture: 'user' | 'environment' | string | null | undefined;
```
jeripeierSBB, line 236: Here I am also not that sure, because types here are not the same as they are here (string missing): 
```ts
/** Attribute whose state specifies the preferred facing mode for the media capture mechanism. */
  @Input() capture?: 'user' | 'environment';
```

src/angular/file-selector/file-selector.ts
```diff
@@ -232,9 +232,4 @@ export class SbbFileSelector
...
-  static ngAcceptInputType_multipleMode: 'default' | 'persistent' | string | null | undefined;
```
jeripeierSBB, line 237: string / null / undefined?
```ts
@Input() multipleMode: 'default' | 'persistent' = 'default';
```

src/angular/table/table/text-column.ts
```diff
@@ -49,8 +49,8 @@ export class SbbTextColumn<T> extends CdkTextColumn<T> implements OnInit {
   get groupWithNext(): boolean {
     return this._groupWithNext;
   }
-  set groupWithNext(value: boolean) {
-    this._groupWithNext = value;
+  set groupWithNext(value: BooleanInput) {
+    this._groupWithNext = coerceBooleanProperty(value);
```
jeripeierSBB, line 53: nice catch :-)

jeripeierSBB (commented):
Why don't we need to set types in mixins? e.g. tabindex.ts there is still number as type. (but also in angular/components, so it seems to be correct).
A few other questions on which you probably will have an answer :)

PR #1023, opened 2022-01-14
fix(angular/pagination): add page index checks to change page methods

kyubisation:


jeripeierSBB (changes requested):
I think the newly added logic was already covered by our pageIndex setter. Maybe pageIndex++ did not call the setter? Additionally, if this really fixes a case, I think a unit test should be added.

PR #1027, opened 2022-01-17
fix(angular/core): make SbbOption generic

jeripeierSBB:
Makes `SbbOption` generic so that consumers can type its `value` to something different from `any`.
https://github.com/angular/components/pull/20242


kyubisation (approved):
LGTM 👍

PR #1028, opened 2022-01-17
refactor(angular/input): inconsistently reading name from input with ngModel

jeripeierSBB:
If an input has a `name` binding and an `ngModel`, the input harness won't be able to
read the name from the DOM, because `ngModel` doesn't proxy it. These changes add
the proxy behavior to the `SbbInput` directive, similarly to what we we're doing for
`required`, `placeholder`, `readonly` etc.
https://github.com/angular/components/pull/19233


kyubisation (approved):
LGTM 👍 
`MatInput` is used in the commit message

PR #1029, opened 2022-01-17
fix(angular/autocomplete): don't handle enter events with modifier keys

jeripeierSBB:
Doesn't handle `ENTER` key presses and doesn't prevent their default action,
if they have a modifier, in order to avoid interfering with any OS-level shortcuts.
https://github.com/angular/components/pull/14717


kyubisation (approved):
LGTM 👍

PR #1030, opened 2022-01-17
fix(angular/table): better handling of invalid data

jeripeierSBB:
The table data source is set up to expect an array and will throw a cryptic error
 down the line if the value is anything different. While typings should be enough to enforce this,
 if the value comes from somewhere in the view it may not get caught.
Since the effort for handling it on our end is minimal,
these changes add some logic that fall back to an empty array if the value is invalid.
https://github.com/angular/components/pull/18953


kyubisation (approved):
LGTM 👍

PR #1031, opened 2022-01-17
fix(angular/sidebar): implicit content element being registered twice with scroll dispatcher

jeripeierSBB:
When the consumer doesn't provide a `sbb-sidebar-content`, we create one implicitly for them.
The implicit content element has a `cdkScrollable` on it, which means that it'll be
registered on the `ScrollDispatcher` as a `CdkScrollable`, however since `SbbSidebarContent`
also extends `CdkScrollable`, it'll be registered again as a `SbbSidebarContent`.
These change remove the extra `cdkScrollable` from the view in order to avoid the issue.

These changes also provide the sidenav content as a `CdkScrollable` for DI purposes.
https://github.com/angular/components/pull/13973


kyubisation (approved):
LGTM 👍

PR #1033, opened 2022-01-17
fix(angular/autocomplete): optionSelections not emitting when the list of options changes

jeripeierSBB:
Fixes the `SbbAutocompleteTrigger.optionSelections` stopping to emit when
the list options has been swapped out.
https://github.com/angular/components/issues/14777


kyubisation (approved):
LGTM 👍

PR #1034, opened 2022-01-17
fix(angular/tooltip): not closing if escape is pressed while trigger isn't focused

jeripeierSBB:
Currently, the tooltip's `keydown` handler is on the trigger, which means that it won't fire
if the trigger doesn't have focus. This could happen when a tooltip was opened by
hovering over the trigger. These changes use the `OverlayKeyboardDispatcher` to
handle the events instead.
https://github.com/angular/components/pull/14434


kyubisation (approved):
LGTM 👍

PR #1035, opened 2022-01-17
fix(angular/chips): declare that SbbChipInput.inputElement is always defined

jeripeierSBB:
It is initialized in the constructor, so there is no chance of it being
undefined.
https://github.com/angular/components/pull/23868


kyubisation (approved):
LGTM 👍

PR #1037, opened 2022-01-17
fix(angular/badge): correctly apply badge description

jeripeierSBB:
Previously, `SbbBadge` was applying the provided description
as an `aria-label` to the badge content element, which is
a child element of the badge host. Then it was _also_ applying
an `aria-describedby` to that same content element.
The end result is that the badge was being treated
mainly as text content for a control, which isn't really
the best way to convey its complementary nature.
This behavior was causing problems in NVDA,
where the badge content was overriding the host button's label.

This change makes the badge content itself `aria-hidden` and
applies the badge description to the badge's _host_ element
via `aria-describedby` and completely removes any setting of `aria-label`.
It also makes a handful of minor cleanup refactorings.

https://github.com/angular/components/pull/23562


kyubisation (approved):
LGTM 👍

PR #1042, opened 2022-01-17
fix(angular/sidebar): restore focus with correct origin when closing via the backdrop

jeripeierSBB:
Currently, when a sidebar is closed, we restore the focus origin as when it was focused
because we don't know what kind of event closed it, however if it happens through
the backdrop, we can be fairly certain that it was via mouse.

These changes pass in the `mouse` origin when closing through the backdrop.

https://github.com/angular/components/pull/23492


kyubisation (approved):
LGTM 👍

PR #1043, opened 2022-01-17
fix(angular/autocomplete): don't emit optionActivated event when option is reset

jeripeierSBB:
Fixes that we were emitting the `optionActivated` event when the option is
reset programmatically when the panel closes.
The intent was to only emit the event as a result of a user action.

https://github.com/angular/components/pull/23437


kyubisation (approved):
LGTM 👍

PR #1044, opened 2022-01-17
fix(angular/icon): make icon-registry compatible with Trusted Types

jeripeierSBB:
When SBB Angular is used in an environment that enforces Trusted
Types, the icon registry raises a Trusted Types violation due to its use
of element.innerHTML when initializing SVG icons.

To make the icon registry compatible with Trusted Types,
SvgIconConfig.svgText is changed to a TrustedHTML, and its users updated
to either produce TrustedHTML (making sure to only do so in cases where
its security can be readily assessed) or pass such values along.

To facilitate this, add a module that provides a Trusted Types policy,
'sbb-esta#angular'. The policy is created lazily and stored in a
module-local variable. This is the same as the approach taken by Angular
proper in
https://github.com/angular/angular/blob/master/packages/core/src/util/security/trusted_types.ts


kyubisation (approved):
LGTM 👍

PR #1046, opened 2022-01-17
fix(angular/select): component value not in sync with control value on init

jeripeierSBB:
Fixes the `SbbSelect.value` property not being in sync
with the `ControlValueAccessor` initial value on init.

https://github.com/angular/components/pull/18443


kyubisation (approved):
LGTM 👍

PR #1057, opened 2022-01-19
fix(angular/tooltip): fix displaying close icon if click trigger was used

### LINKED ISSUE #1050: Bug: tooltip with template and trigger "click" is missing close button
**Describe the bug**
when creating a tooltip with a template and the trigger "click" there is no close button:
image

jeripeierSBB:
Closes #1050


src/angular/tooltip/tooltip.ts
```diff
@@ -818,6 +820,9 @@ export abstract class _TooltipComponentBase implements OnDestroy {
   /** Subject for notifying that the tooltip has been hidden from the view */
   private readonly _onHide: Subject<void> = new Subject<void>();
 
+  /** Whether the tooltip component was triggered by click */
+  triggeredByClick: boolean = false;
```
kyubisation, line 824: This should be prefixed with _

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #1068, opened 2022-01-24
fix(angular/autocomplete): restore focus after emitting option selected event

jeripeierSBB:
Currently, we restore focus to the input and then we emit the change event, but we have a report that it may be making some use cases more difficult. From what I can tell, this shouldn't have much of an impact on existing users so these changes swap the order so that the focus event is last.

https://github.com/angular/components/commit/7409bd615f63a0a5d2c2ccee9ce278b0cce6bb69


kyubisation (approved):
LGTM 👍

PR #1080, opened 2022-01-26
fix(angular/radio-button-panel): only show error styles when parent group is invalid

### LINKED ISSUE #1078: Bug: Radio-Button-Panel with invalid indication when form is touched, but input is not touched
**Describe the bug**
The radio button panel (with required validation) is shown with red border, even though it is untouched. 
This is probably caused by a too wide selector, that matches the outer form touched state.

**To Reproduce**
1. nest a required radio button panel in a form
2. edit any other field in that form
3. observe that the radio button panel gets red borders

**Expected behavior**
red borders should only occur when the actual input is touched.

**Screenshots**
image

 - OS: Win10
 - Browser Chrome
 - Version SACL 13.2.0

**Additional context**
This might affect other inputs as well - I did not check this exhaustively..

kyubisation:
Previously we would show error styles, when any ancestor had .ng-touched.ng-invalid assigned.
This could also be the FormGroup element, which would cause the radio-button-panel
to incorrectly be marked as invalid.

Closes #1078

### DIFF
--- a/src/angular/radio-button/radio-button.ts
+++ b/src/angular/radio-button/radio-button.ts
@@ -67,7 +67,11 @@ export const SBB_RADIO_GROUP = new InjectionToken<_SbbRadioGroupBase<_SbbRadioBu
  */
 export const SBB_RADIO_BUTTON = new InjectionToken<_SbbRadioButtonBase>('SbbRadioButton');
 
-@Directive()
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM

PR #1090, opened 2022-01-27
fix(angular/tabs): fix showing pagination arrows on zoom levels other than 100%

jeripeierSBB:


src/angular/tabs/paginated-tab-header.ts
```diff
@@ -493,7 +493,8 @@ export abstract class SbbPaginatedTabHeader
       this._showPaginationControls = false;
     } else {
       const isEnabled =
-        this._tabListInner.nativeElement.scrollWidth > this._elementRef.nativeElement.offsetWidth;
+        this._tabListInner.nativeElement.scrollWidth - this._elementRef.nativeElement.offsetWidth >
+        1;
```
jeripeierSBB, line 497:
```suggestion
        1; // Allow difference delta of 1 to avoid falsy enabled pagination if browser has zoom levels other than 100%
```

src/angular/tabs/paginated-tab-header.ts
```diff
@@ -493,7 +493,8 @@ export abstract class SbbPaginatedTabHeader
       this._showPaginationControls = false;
     } else {
       const isEnabled =
-        this._tabListInner.nativeElement.scrollWidth > this._elementRef.nativeElement.offsetWidth;
+        this._tabListInner.nativeElement.scrollWidth - this._elementRef.nativeElement.offsetWidth >
+        1; // Allow difference delta of 1 to avoid falsy enabled pagination if browser has zoom levels other than 100%
```
kyubisation, line 497: Nit; Place comments before the described part to avoid prettier freaking out. (Additionally from my point of view it's easier to read; have the explenation before the affected code part.)

kyubisation (approved):
LGTM 👍

PR #1091, opened 2022-01-28
docs(table): correct closing tag

julienje:
On some examples, the closing tag for sbb-table-wrapper is a div instead of sbb-table-wrapper


kyubisation (approved):
LGTM 👍 
Thanks for the catch! 😃

PR #1094, opened 2022-02-01
feat(angular/tabs): add the ability to keep content inside the DOM while off-screen

jeripeierSBB:
Adds the `preserveContent` input which allows consumers to opt into keeping the content of off-screen tabs inside the DOM. This is useful primarily for edge cases like iframes and videos where removing the element from the DOM will cause it to reload.

One gotcha here is that we have to set `visibility: hidden` on the off-screen content so that users can't tab into it.

https://github.com/angular/components/commit/cad087220848b2cd2c252520bac26a42d8b279aa

### DIFF
--- a/src/angular/tabs/tab-body.scss
+++ b/src/angular/tabs/tab-body.scss
@@ -5,4 +5,15 @@
   .sbb-tab-group-dynamic-height & {
     overflow: hidden;
   }
+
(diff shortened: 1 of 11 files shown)


kyubisation (approved):
LGTM 👍

PR #1102, opened 2022-02-02
fix(angular/tabs): use buttons for paginator

jeripeierSBB:
Uses `button` elements, rather than styled `div`-s for the paginator buttons. This has the advantage of being more accessible, if we decided to make them focusable, and it stops calling the click listeners when the button is disabled (currently we do some expensive checks in them when we don't have to).


kyubisation (approved):
LGTM 👍

PR #1103, opened 2022-02-03
fix(angular/menu): adjust overlay size when amount of items changes

jeripeierSBB:
Currently, we lock the menu into a position after it is opened so that it doesn't jump around when the user scrolls, but this means that if the amount of items changes, it might not be the optimal position anymore. These changes add some code to re-calculate the position if the amount of items changes.


src/angular/menu/menu-item.ts
```diff
@@ -136,10 +133,18 @@ export class SbbMenuItem
...
-      icon.parentNode?.removeChild(icon);
+      icons[i].remove();
     }
 
     return clone.textContent?.trim() || '';
   }
+
+  _setHighlighted(isHighlighted: boolean) {
+    // We need to mark this for check for the case where the content is coming from a
+    // `matMenuContent` whose change detection tree is at the declaration position,
```
kyubisation, line 144: Do we have a sbbMenuContent?

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #1104, opened 2022-02-03
fix(angular/datepicker): make input full width

mhaertwig:
This bug only appears when the datepicker is not inside an `sbb-form-field`. Therefore I've added an test to ensure it is full width.

### DIFF
--- a/src/angular/datepicker/datepicker/datepicker.scss
+++ b/src/angular/datepicker/datepicker/datepicker.scss
@@ -12,6 +12,8 @@
 }
 
 .sbb-date-input {
+  width: 100%;
+
   :is(.sbb-datepicker-toggle-enabled, .sbb-datepicker-arrows-enabled) > & {
     padding-left: var(--sbb-datepicker-toggle-or-arrows-padding);
   }
--- a/src/angular/datepicker/datepicker/datepicker.scss
+++ b/src/angular/datepicker/datepicker/datepicker.scss
@@ -5,7 +5,7 @@
   position: relative;
   color: var(--sbb-color-weak-accent);
 
-  .sbb-form-field-input-wrapper > & {
+  sbb-form-field & {
     width: 100%;
   }
 }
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM, thanks!

src/angular/datepicker/datepicker/datepicker.scss
```diff
@@ -12,6 +12,8 @@
 }
 
 .sbb-date-input {
+  width: 100%;
```
kyubisation, line 15:
```suggestion
  .sbb-datepicker > & {
    width: 100%;
  }
```

kyubisation (changes requested):
Also remove the 100% from .sbb-datepicker

src/angular/datepicker/datepicker/datepicker.scss
```diff
@@ -3,15 +3,22 @@
 .sbb-datepicker {
   display: inline-block;
   position: relative;
-  width: 100%;
   color: var(--sbb-color-weak-accent);
+
+  .sbb-form-field-input-wrapper > & {
```
jeripeierSBB, line 8: A little nit picking, but i think form-field-input-wrapper is more like an implementation detail of form field. @kyubisation your thoughts?
```suggestion
  .sbb-form-field > & {
```

mhaertwig: I like it since it's more exlicit. I committed it without the `>` since `sbb-form-field` is not a direct parent.

jeripeierSBB (approved):
LGTM

src/angular/datepicker/datepicker/datepicker.scss
kyubisation: I agree. However, the `>` operator should be removed in that case.

src/angular/datepicker/datepicker/datepicker.scss
```diff
@@ -3,15 +3,22 @@
 .sbb-datepicker {
   display: inline-block;
   position: relative;
-  width: 100%;
   color: var(--sbb-color-weak-accent);
+
+  sbb-form-field & {
```
kyubisation, line 8: Nitpick (again); Use class selector unless there is a specific reason (like specificity).
```suggestion
  .sbb-form-field & {
```

kyubisation (approved):
LGTM 👍

PR #1105, opened 2022-02-03
fix(angular/datepicker): change calendar cells to buttons

jeripeierSBB:
Makes changes to the DOM structure of calendar cells for better screen reader experience. Previously, the DOM
structure looked like this:

```
<!-- Existing DOM structure of each calendar body cell -->
<td
 class="sbb-calendar-body-cell"
 role="gridcell"
 aria-disabled="false"
 aria-current="date"
 aria-selected="true"
 <!-- ... -->
>
 <!-- additional details ommited -->
</>
```

Using the `gridcell` role allows screenreaders to use table specific
navigation and some screenreaders would announce that the cells are
interactible because of the presence of `aria-selected`. However, some
screenreaders did not announce the cells as interactable and treated it
the same as a cell in a static table (e.g. VoiceOver announces element
type incorrectly angular/components#23476).

This changes the DOM structure to nest buttons
inside of a gridcell to make it more explicit that the table cells can
be interacted with and are not static content. The gridcell role is
still present, so table navigation will continue to work, but the
interaction is done with buttons nested inside the `td` elements.
The `td` element is only for adding information to the a11y tree and not used for visual purposes.

Updated DOM structure:

```
<td
  role="gridcell"
  class="sbb-calendar-body-cell-container"
>
  <button
   class="sbb-calendar-body-cell"
   aria-disabled="false"
   aria-current="date"
   aria-pressed="true"
   <!-- ... -->
  >
   <!-- additional details ommited -->
  </button>
</td>
```

Signed-off-by: Jeremias Peier <jeremias.peier@sbb.ch>

### DIFF
--- a/src/angular/datepicker/calendar-body/calendar-body.html
+++ b/src/angular/datepicker/calendar-body/calendar-body.html
@@ -29,7 +29,7 @@
       [class.sbb-calendar-body-disabled]="!item.enabled"
       [class.sbb-calendar-body-active]="isActiveCell(rowIndex, colIndex)"
       [class.sbb-calendar-body-range-background]="item.rangeBackground === 'range'"
-      [class.sbb-calendar-body-range-background-offset-start]="item.rangeBackground === 'range' && rowIndex === 0 && firstRowOffset && colIndex === 0"
+      [class.sbb-calendar-body-range-background-offset-start]="item.rangeBackground !== null && rowIndex === 0 && firstRowOffset && colIndex === 0"
       [class.sbb-calendar-body-selected-begin]="item.rangeBackground === 'begin'"
       [class.sbb-calendar-body-selected-end]="item.rangeBackground === 'end'"
       [class.sbb-calendar-body-selected]="selectedValue === item.value"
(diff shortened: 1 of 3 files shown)


kyubisation:
I think I figured out what the problem is.
https://github.com/sbb-design-systems/sbb-angular/blob/e4c6cc47d2048b7dc2632115b0abf7bc9af06430/src/angular/datepicker/datepicker-content/datepicker-content.scss#L257-L266
This rule needs to be adapted to the changed circumstances of the classes being nested.

The following code might work:
```scss
  // If there is only one label in a row, or start is first or last entry
  :is(.sbb-calendar-body-cell-container:first-child:last-child
      > &.sbb-calendar-body-range-background, .sbb-calendar-body-cell-container:last-child
      > &.sbb-calendar-body-range-background-offset-start, .sbb-calendar-body-cell-container:first-child
      > &.sbb-calendar-body-selected-end, .sbb-calendar-body-cell-container:last-child
      > &.sbb-calendar-body-selected-begin) {
    &::before {
      display: none;
    }
    &::after {
      border-radius: 50%;
    }
  }
```

kyubisation (approved):
LGTM 👍 
Nice work 😃

PR #1110, opened 2022-02-08
fix(angular): fix migration for typescript versions smaller or equals 4.4

jeripeierSBB:


src/angular/schematics/ng-add/migrations/secondary-entry-points-migration.ts
```diff
@@ -255,9 +258,9 @@ function applyRenames(elements: ts.ImportSpecifier[]) {
         }
 
         const nameIdentifier = ts.createIdentifier(CLASS_NAME_RENAMES.get(elementName.text)!);
-        return element.propertyName
-          ? ts.createImportSpecifier(false, nameIdentifier, element.name)
-          : ts.createImportSpecifier(false, undefined, nameIdentifier);
+        return elementName
```
jeripeierSBB, line 261:
```suggestion
        return element.propertyName
```

kyubisation (approved):
LGTM 👍

PR #1113, opened 2022-02-08
test: clean up IE/Edge test workarounds

kyubisation:
The workarounds shouldn't be necessary since we don't run tests against IE and legacy Edge anymore.


jeripeierSBB (approved):
LGTM

Why the yarn lock change?

kyubisation:
I'm honestly just as confused as you are. But running yarn install added this change.

PR #1114, opened 2022-02-08
fix(angular): allow coercing of booleans for all inputs

jeripeierSBB:
Previously, some properties were declared as BooleanInput and some not. With this commit, all boolean inputs are threat as BooleanInput and therefore the input values are being coerced.


kyubisation (approved):
LGTM 👍

PR #1118, opened 2022-02-09
fix(angular/datepicker): update active date on focusing a calendar cell

jeripeierSBB:
When a date cell on the calendar recieves focus, set the active date to that cell. This ensures that the active date matches the date with browser focus.

Previously, we set the active date on keydown and click, but that was problematic for screenreaders. That's because many screenreaders trigger a focus event instead of a keydown event when using screenreader specific navigation (VoiceOver, Chromevox, NVDA).


jeripeierSBB:
waiting for resubmit in #angular/components due to reverted commit: https://github.com/angular/components/commit/e04e4a781b9d7721b70afe69d1da8336398901a0

kyubisation (approved):
LGTM 👍

PR #1120, opened 2022-02-09
fix(angular/datepicker): fix improper focus trapping with VoiceOver and ChromeVox

jeripeierSBB:
Fixes focus trapping on the datepicker dialog by putting `role="dialog"` `aria-modal="true"` and `cdkTrapFocus` all on the same element. This aligns the datepicker with how SbbDialog does focus trapping.

Without having them all on the same element, users could exit out of the focus trapping using screenreader specific navigation with VoiceOver and ChromeVox.

### DIFF
--- a/src/angular/datepicker/datepicker-content/datepicker-content.html
+++ b/src/angular/datepicker/datepicker-content/datepicker-content.html
@@ -1,5 +1,9 @@
 <sbb-calendar
   cdkTrapFocus
+  role="dialog"
+  [attr.aria-modal]="true"
+  [attr.aria-labelledby]="_dialogLabelId ?? undefined"
+  class="sbb-datepicker-content-container"
   [id]="datepicker.id"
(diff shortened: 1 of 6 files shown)


src/angular/datepicker/datepicker-content/datepicker-content.html
```diff
@@ -1,5 +1,9 @@
 <sbb-calendar
   cdkTrapFocus
+  role="dialog"
+  [attr.aria-modal]="true"
```
kyubisation, line 4: Is this intentionally not `aria-modal="true"`?

kyubisation (approved):
LGTM 👍 
One question about an attribute

src/angular/datepicker/datepicker-content/datepicker-content.html
jeripeierSBB: hmm copied from material

PR #1122, opened 2022-02-10
fix(angular/toggle): fix toggle animation

### LINKED ISSUE #652: Bug: sbb-toggle animation of content disabled
**Describe the bug**
There is no animation when switching between toggle options if they have content.

**Expected behavior**
Content of an sbb-toggle-option should be animated when changing between options (fade in / out and changing height).

**Additional context**
Because of the new DOM structure introduced in #651 it was not easy to adapt the animation. Needs further investigation.

jeripeierSBB:
Closes #652


kyubisation (changes requested):
Nice work!
One small issue.

kyubisation (commented):
LGTM 👍

PR #1129, opened 2022-02-14
fix(angular/table): fix line-height of standard design variant

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
(I'm a bit surprised about this, but seems correct 😅 )

PR #1135, opened 2022-02-15
refactor(angular/sidebar): remove PlatformModule usage

jeripeierSBB:
The `PlatformModule` is a noop, so we don't need to import it.


kyubisation (approved):
LGTM 👍

PR #1136, opened 2022-02-15
fix(angular/menu): position classes not updated when window is resized

jeripeierSBB:
We were calling `SbbMenu.setPositionClasses` in order to set the classes if the position changes, but the call was outside of the `NgZone` so nothing was happening. These changes bring the call back into the `NgZone`.


kyubisation (approved):
LGTM 👍

PR #1137, opened 2022-02-15
fix(angular/chips): update chip-list describedby to match input

jeripeierSBB:
Fixes ExpressionChangedAfterItHasBeenCheckedError that occurs when chip-lists are content-projected into a sbb-form-field


kyubisation (approved):
LGTM 👍

PR #1140, opened 2022-02-16
feat(angular/table): allow hover action buttons

jeripeierSBB:


src/angular/table/table.md
```diff
@@ -477,6 +477,36 @@ The flag affects the sequence at runtime, not at definition time.
 </ng-container>
 ```
 
+### Table Cell Actions
+
+It is possible to define up to three action buttons (or links) for one row.
```
kyubisation, line 482: Is there a specific reason/limitation as to why we only support up to three?

src/angular/table/table.md
```diff
@@ -477,6 +477,36 @@ The flag affects the sequence at runtime, not at definition time.
...
+not immediately see that there are actions.
+We recommend for accessibility reasons to always create a click action on the whole row,
+which links to a possible detail page, where all actions can be called again.
+Also, the action buttons should always be provided with aria labels,
+so that screen readers can output a meaningful text.
+
+```html
+<td>
+  <div class="sbb-table-cell-actions" aria-label="Press tab key to access actions.">
+    <a sbb-icon-button href="#" attr.aria-label="Edit {{element.vehicle}}.">
```
kyubisation, line 496:
```suggestion
    <a sbb-icon-button [routerLink]="['./edit', element.id]" attr.aria-label="Edit {{element.vehicle}}.">
```

src/angular/table/table.md
jeripeierSBB: It comes from design specification. But we don‘t need to mention it here. What do you think?

src/angular/table/table.md
jeripeierSBB: But we leave it in the example, or what do you think?

src/angular/table/table.md
jeripeierSBB, line 482:
```suggestion
It is possible to define up to three (recommended, more possible) action buttons (or links) for one row.
```

kyubisation (approved):
LGTM 👍

PR #1142, opened 2022-02-16
fix(angular/lightbox): hide scrollbar from scroll lock container

### LINKED ISSUE #1139: Bug: Lightbox does not have the "correct" scrollbar
**Describe the bug**
In our application we open a Lightbox with some form-content. We can add form field groups dynamically. When we have an overflow, we can scroll with the mouse wheel and keyboard, but the scrollbar is fix with no scroll at all.

**To Reproduce**
1. Go to ['Webshop Offer Test Environment'](https://webshop-offers-web-snapshot.app.sbb.ch/de/ip-offers) - Lightbox opens automatically
2. Click on 'Reisende Hinzufügen' until container overflow
4. See error (no "correct" scrollbar in right side)

**Expected behavior**
We expect the lightbox scrollbar as shown in https://angular.app.sbb.ch/angular/components/lightbox/examples

**Screenshots**
Not Good:
image

 - OS: Windows
 - Browser Edge,...

**Additional context**
Played around with stylings. When lightbox is opened, the html element has the class .cdk-global-scrollblock with a rule overflow-y: scroll;. When we remove this rule, everything works fine:
image

kyubisation:
Closes #1139

### DIFF
--- a/src/angular/styles/_typography.scss
+++ b/src/angular/styles/_typography.scss
@@ -592,6 +592,12 @@ textarea {
   max-height: none;
 }
 
+// Overwrites the 'scroll' value from the CDK, since this causes a scrollbar to be shown
+// on top of the lightbox.
+.cdk-global-scrollblock {
+  overflow-y: auto;
+}
+
 // ----------------------------------------------------------------------------------------------------
 // Headings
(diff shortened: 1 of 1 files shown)


jeripeierSBB (approved):
LGTM

PR #1153, opened 2022-02-17
fix(angular/lightbox): fix aria-label on close button

### LINKED ISSUE #1152: Bug: no aria-label for close icon button in lightbox header
**Describe the bug**
We got feedback in terms of accessibility for our application. One thing was that the close button in the lightbox header does not have an alt text or aria label. Since it is necessary for icon-only buttons, we should have a configuration property or a default text for that, e.g. "close the dialog"

**To Reproduce**
1. Go to ['Ligbtox Examples page'](https://angular.app.sbb.ch/angular/components/lightbox/examples)
2. Click on 'Open Lightbox from component'
3. Inspect close button in upper right corner with accessibility viewer or screen reader
4. See no aria label or description

**Expected behavior**
Should have a default aria label or a config to set this

**Screenshots**
image

jeripeierSBB:
Closes #1152

### DIFF
--- a/src/angular/lightbox/lightbox-content-directives.ts
+++ b/src/angular/lightbox/lightbox-content-directives.ts
@@ -2,6 +2,7 @@
 /// <reference types="@angular/localize/init" />
 
 import {
+  AfterViewInit,
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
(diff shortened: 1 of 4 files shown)


src/angular/lightbox/lightbox-content-directives.ts
```diff
@@ -95,10 +89,12 @@ export class SbbLightboxClose extends SbbDialogClose implements OnInit, OnChange
     '[id]': 'id',
   },
 })
-export class SbbLightboxTitle extends _SbbDialogTitleBase {
+export class SbbLightboxTitle extends _SbbDialogTitleBase implements AfterViewInit {
   /** Unique id for the lightbox title. If none is supplied, it will be auto-generated. */
   @Input() override id: string = `sbb-lightbox-title-${dialogElementUid++}`;
 
+  @ViewChild(SbbLightboxClose) _lightBoxClose: SbbLightboxClose;
```
kyubisation, line 96: It should be possible to load this statically?

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

PR #1170, opened 2022-02-21
build: use less ambiguous peer dependency to Angular framework

jeripeierSBB:


kyubisation (approved):
I'm not quite sure, why the Bazel linter fails. Maybe we have to exclude the .git directory?

PR #1172, opened 2022-02-21
fix(angular/radio-button): not checked on first click if partially visible

jeripeierSBB:
Currently, the native `input` inside the radio button is collapsed down to 1px by 1px
and moved to the bottom of the element which causes browsers to scroll the window
down, rather than change the value, on the first click when the input is partially
hidden. These changes fix the issue by having the input cover the entire button and
hiding it with `opacity`. From my testing, using `opacity` versus `cdk-visually-hidden`
doesn't make a difference for screen readers.


kyubisation (approved):
This prevents the text in the radio button from being selected, but I'm not sure whether that is necessary, so I think it should be fine.

jeripeierSBB:
> This prevents the text in the radio button from being selected, but I'm not sure whether that is necessary, so I think it should be fine.

Fixed by #1178

PR #1176, opened 2022-02-22
feat(angular/lightbox): allow aria-label of close button to customize

jeripeierSBB:


src/angular/dialog/dialog-content-directives.ts
```diff
@@ -21,6 +21,11 @@ import { SbbDialogRef, _closeDialogVia } from './dialog-ref';
 /** Counter used to generate unique IDs for dialog elements. */
 let dialogElementUid = 0;
 
+const closeAriaLabel = () =>
```
kyubisation, line 24: I'd prefer a constant instead of a "function". Or was there a reason against it?

src/angular/dialog/dialog-content-directives.ts
```diff
@@ -107,6 +109,9 @@ export class _SbbDialogTitleBase implements OnInit {
   /** Unique id for the dialog title. If none is supplied, it will be auto-generated. */
   @Input() id: string = `sbb-dialog-title-${dialogElementUid++}`;
 
+  /** Arial label for the close button. */
+  @Input() closeArialabel: string = closeAriaLabel();
```
kyubisation, line 113: I think we used `closeAriaLabel` casing in other places.

src/angular/lightbox/lightbox-content-directives.ts
```diff
@@ -22,6 +22,11 @@ import { SbbLightboxRef } from './lightbox-ref';
 /** Counter used to generate unique IDs for dialog elements. */
 let dialogElementUid = 0;
 
+const closeAriaLabel = () =>
```
kyubisation, line 25: Same consideration as with dialog. I'd prefer a constant.

kyubisation (changes requested):

PR #1177, opened 2022-02-22
docs(angular/lightbox): remove obsolete button alignment infos

jeripeierSBB:


kyubisation (changes requested):
As far as I know, the specification of lightbox is stricter in regard to button placement.
I don't know whether we should support this?

src/angular/lightbox/lightbox-content-directives.ts
```diff
@@ -132,6 +132,16 @@ export class SbbLightboxContent {}
...
+    '[class.sbb-lightbox-actions-align-center]': 'align === "center"',
+    '[class.sbb-lightbox-actions-align-end]': 'align === "end"',
+  },
 })
-export class SbbLightboxActions {}
+export class SbbLightboxActions {
+  /**
+   * Horizontal alignment of action buttons.
+   */
+  @Input() align?: 'start' | 'center' | 'end' | 'space between' = 'space between';
```
mhaertwig, line 146: Do we need the question mark after `align` if there's a default value?
I would prefer `space-between` (with `-`).

jeripeierSBB:
@mhaertwig We decided to not introduce this feature, but make a doc fix instead.

kyubisation (approved):
LGTM 👍

mhaertwig:
LGTM 👍

PR #1178, opened 2022-02-22
fix(angular/radio-button): clicks not propagating to wrapper elements

jeripeierSBB:
Fixes a regression caused by #1172 where clicking directly on the `input` element will prevent the click from reaching custom `click` listeners outside the radio button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -239,6 +239,11 @@
         0,
         "always",
         0
+      ],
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍

mhaertwig:
LGTM 👍

PR #1189, opened 2022-02-22
feat(angular/header-lean): align icons in app-chooser-section

mhaertwig:


src/angular/header-lean/app-chooser-section.scss
```diff
@@ -47,4 +47,20 @@
   &:last-child {
     border-bottom: var(--sbb-border-width-thin) solid var(--sbb-color-cloud);
   }
+
+  > :is(.sbb-icon, img) {
```
jeripeierSBB, line 51: I'm not sure whether we should use '>' or not. I think an icon inside an icon will never happen, but maybe someone uses a span inside the `<a>` or whatever. What do you think?

jeripeierSBB (commented):
Doesn't it make the size of a link entry bigger like it is implemented? I expected that some minus margin is necessary?

src/angular/header-lean/app-chooser-section.scss
mhaertwig: Makes sense if we want to allow child elements

src/angular/header-lean/app-chooser-section.scss
```diff
@@ -47,4 +47,25 @@
   &:last-child {
     border-bottom: var(--sbb-border-width-thin) solid var(--sbb-color-cloud);
   }
+
+  .sbb-icon,
+  img {
```
jeripeierSBB, line 52:
```suggestion
  :is(.sbb-icon, img) {
```

jeripeierSBB: should be :is here, because it's nested in `a` and it would generate duplicated definitions on sass compilation.

src/angular/header-lean/app-chooser-section.scss
```diff
@@ -47,4 +47,25 @@
...
+    vertical-align: middle;
+    width: var(--sbb-icon-size-default);
+    height: var(--sbb-icon-size-default);
+
+    // adjust margin to keep line height
```
jeripeierSBB, line 57:
```suggestion
    // Adjust margin to keep line height
```
We once started to write comment starting letters upper case

src/angular/header-lean/app-chooser-section.scss
```diff
@@ -47,4 +47,25 @@
...
+    margin-top: calc((var(--sbb-line-height-medium) - var(--sbb-icon-size-default)) / 2);
+    margin-bottom: calc((var(--sbb-line-height-medium) - var(--sbb-icon-size-default)) / 2);
+
+    svg {
+      width: 100%;
+      height: 100%;
+    }
+
+    &:first-child {
+      float: left; // used to ignore users whitespace and to ensure margin is correct
```
jeripeierSBB, line 67:
```suggestion
      float: left; // Used to ignore users whitespace and to ensure margin is correct
```
We once started to write comment starting letters upper case

jeripeierSBB (changes requested):

src/angular/header-lean/app-chooser-section.scss
mhaertwig: Good to know. However, I've removed the float after our current discussion.

jeripeierSBB (approved):
LGTM

PR #1205, opened 2022-02-24
fix(angular/autocomplete): use narrow value for aria-haspopup

jeripeierSBB:
[Based on the spec](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup),
`aria-haspopup` can be set to indicate what kind of popup the element has.
These changes update the autocomplete one to show that it opens a listbox.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM

PR #1208, opened 2022-02-24
fix(angular/autocomplete): closing immediately when input is focused programmatically

jeripeierSBB:
Each autocomplete has a `click` listener on the body that closes the panel when the
user has clicked somewhere outside of it. This breaks down if the panel is opened
through a click outside of the form field, because we bind the listener before the
event has bubbled all the way to the `body` so when it does get there, it closes
immediately.

These changes fix the issue by taking advantage of the fact that focus usually moves
on `mousedown` so for "real" click the input will have already lost focus by the time
the `click` event happens.


mhaertwig (approved):
LGTM 👍

kyubisation (approved):
LGTM 👍

PR #1209, opened 2022-02-24
fix(angular/sidebar): prevent focus from entering hidden sidenav if child element has a visibility

jeripeierSBB:
This is a similar fix that we've made for other components. Usually we set `visibility: hidden` on the sidebar which is enough to prevent the user from tabbing into it, however one of the children could override it with `visibility: visible`. These changes add an extra `display: none` on top which will prevent such issues. This has come up on AIO recently angular/angular#44990.


mhaertwig (approved):
LGTM 👍

kyubisation (approved):
LGTM 👍

PR #1210, opened 2022-02-24
fix(angular/checkbox): clear static aria attributes from host nodes

jeripeierSBB:
Clears the aria-* attributes from the host node so that they're not
duplicated with the underlying input.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1211, opened 2022-02-24
fix(angular/autocomplete): don't block default arrow keys when using modifiers

jeripeierSBB:
Currently, we hijack all up/down arrow key events, however this interferes with keyboard shortcuts such as shift + up arrow for marking all of the text. These changes stop intercepting the arrow keys, if they're used with a modifier.

These changes also fix an issue where all the mocked out key events had the `metaKey` set to `true` on some browsers.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1213, opened 2022-02-24
feat(showcase): avoid 404s by only loading available example files

mhaertwig:


src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -1,38 +1,72 @@
...
+  exampleCodes: Observable<ExampleCode[]>;
   showSource = false;
-
+  private _exampleFileOrder = ['html', 'ts', 'css'];
   constructor(private _htmlLoader: HtmlLoader, private _route: ActivatedRoute) {}
 
-  ngOnInit(): void {
-    const exampleName = this.exampleData.selectorName.replace('sbb-', '').replace('-example', '');
+  // Returns the path to the html file for a given example file.
+  private static _convertToHtmlFilePath(filePath: string): string {
```
kyubisation, line 28: We generally don't use static methods. The performance impact is small enough that it usually does not matter.
Also "generally" move private methods to after public methods (there are expections with tight coupling).

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -1,38 +1,72 @@
...
 })
 export class ExampleViewerComponent implements OnInit {
   @Input() exampleData: ExampleData;
-  html: Observable<string>;
-  ts: Observable<string>;
-  css: Observable<string>;
+  exampleCodes: Observable<ExampleCode[]>;
   showSource = false;
-
+  private _exampleFileOrder = ['html', 'ts', 'css'];
```
kyubisation, line 24: We usually have a line separating the constructor from the properties.

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -1,38 +1,72 @@
...
+  private static _convertToExampleName(selectorName: string): string {
+    return selectorName.replace('sbb-', '').replace('-example', '');
+  }
 
-    this.html = this._createLoader(exampleName, 'html');
-    this.ts = this._createLoader(exampleName, 'ts');
-    this.css = this._createLoader(exampleName, 'css');
+  // Get the extension for a given file path
+  private static _getFileExtension(filePath: string): string {
+    return filePath.split('.').pop();
```
kyubisation, line 39: If we have additional files in the future (e.g. for lightbox or dialog) this would still be reduced to .ts or .html. Could we maybe extract the extension from the "default" files and leave it for the rest?

kyubisation (changes requested):
Generally looks good, some small issues

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
mhaertwig: Sorry for missing the lightbox and dialog examples. I've added a filter to the `exampleCodes` observable to ensure that only files whose name match the `indexFileName` (without extension) are displayed.

mhaertwig: I filtered out other example files than the default ones. However, after reading your comment again, I am not sure if I understood you right. Did you want to filter them out or instead display the file names of the non-default tiles.

E.g. titles for lightbox: TS, HTML, CSS, lightbox-example-content.ts

src/showcase/app/shared/component-viewer/example-viewer/example-viewer.component.ts
```diff
@@ -21,58 +21,72 @@ export class ExampleViewerComponent implements OnInit {
...
-        this._createLoader(
-          ExampleViewerComponent._convertToExampleName(this.exampleData.selectorName),
-          exampleFile
-        ).pipe(
-          map((code) => ({
-            label: ExampleViewerComponent._getFileExtension(exampleFile),
-            code,
-          }))
+      this.exampleData.exampleFiles
+        .filter(
```
kyubisation, line 31: Sorry, this was probably a misunderstanding from my wording.
We want to show all example files, but the "default" files should only have the extension as the label. The other example files should be displayed with the "full" file name.

kyubisation (changes requested):

jeripeierSBB (approved):
LGTM, thanks for this great work!

PR #1218, opened 2022-02-25
fix(angular/table): remove role from sort header when disabled

kyubisation:
When the sort header is disabled, we make it non-interactive by clearing its `tabindex`, however it still has the `role="button"` which can throw off screen readers if it doesn't have text.

These changes also clear the `role` from the header if it's disabled.


jeripeierSBB (approved):
LGTM

mhaertwig (approved):
LGTM 👍

PR #1221, opened 2022-02-28
fix(angular/autocomplete): reopen panel on input click

jeripeierSBB:
Currently, if the user clicks an autocomplete to open it, selects an option and then clicks again, the panel won't open, because we use `focus` and the input was focused already. These changes add an extra `click` listener so the panel can reopen.

### DIFF
--- a/src/angular/autocomplete/autocomplete-trigger.ts
+++ b/src/angular/autocomplete/autocomplete-trigger.ts
@@ -20,7 +20,6 @@ import {
   ElementRef,
   forwardRef,
   Host,
-  HostListener,
   Inject,
   InjectionToken,
   Input,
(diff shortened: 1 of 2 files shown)


mhaertwig (approved):
LGTM 👍

src/angular/autocomplete/autocomplete-trigger.ts
```diff
@@ -457,7 +462,6 @@ export class SbbAutocompleteTrigger
   }
 
   /** @docs-private */
-  @HostListener('keydown', ['$event'])
   _handleKeydown(event: TypeRef<KeyboardEvent>): void {
```
kyubisation, line 465: You can remove the TypeRef wrapper as this is only required when using an event type parameter with decorators. (e.g. `TypeRef<KeyboardEvent> => KeyboardEvent`)

kyubisation (changes requested):

PR #1222, opened 2022-02-28
feat(angular/autocomplete): add the ability to auto-select the active option while navigating

### LINKED ISSUE #261: Feature request: autocomplete with auto select first option on blur
To enable faster user input, in our application the first match should automatically be selected when the user leaves an input/autocomplete.

There should be an attribute `autoSelectFirst` (default false) that enables the desired behavior.

**Describe alternatives you've considered**
We are currently working around this, but this is tedious, as the blur event comes before the "selected event".

jeripeierSBB:
Adds the `autoSelectActiveOption` input to `sbb-autocomplete` which allows the
consumer to opt into the behavior where the autocomplete will assign the active
option value as the user is navigating through the list. The value is only propagated
to the model once the panel is closed.

There are a couple of UX differences when the new option is enabled:
1. If the user presses escape while there's a pending auto-selected option, the value
is reverted to the last text they typed before they started navigating.
2. If the user clicks away, tabs away or presses enter while there's a pending option,
it will be selected.

The aforementioned UX differences are based on the Google search autocomplete and
one of the examples from the W3C here:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

Closes #261

### DIFF
--- a/src/angular/autocomplete/autocomplete-trigger.ts
+++ b/src/angular/autocomplete/autocomplete-trigger.ts
@@ -151,6 +151,15 @@ export class SbbAutocompleteTrigger
    */
   private _canOpenOnNextFocus = true;
 
+  /** Value inside the input before we auto-selected an option. */
+  private _valueBeforeAutoSelection: string | undefined;
+
+  /**
+   * Current option that we have auto-selected as the user is navigating,
+   * but which hasn't been propagated to the model value yet.
+   */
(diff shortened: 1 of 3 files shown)


jeripeierSBB:
Is that related to #261?

kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1225, opened 2022-03-01
fix(angular/tooltip): don't hide when pointer moves to tooltip

jeripeierSBB:
Currently, we hide the tooltip as soon as the pointer leaves the trigger element which may be problematic with larger cursors that partially obstruct the content.

These changes allow hover events on the tooltip and add extra logic so that moving to it doesn't start the hiding sequence.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1226, opened 2022-03-01
feat(angular/schematics): add option not to include animations module in ng-add

jeripeierSBB:
Adds a third option to the `ng-add` schematic that allows users to opt out of including any of the animations modules.


kyubisation (approved):
LGTM 👍

PR #1227, opened 2022-03-01
feat(angular/icon): allow multiple classes in setDefaultFontSetClass

jeripeierSBB:
Allows for the consumer to pass in multiple classes to `SbbIconRegistry.setDefaultFontSetClass`.

### DIFF
--- a/src/angular/icon/icon-registry.ts
+++ b/src/angular/icon/icon-registry.ts
@@ -134,10 +134,10 @@ export class SbbIconRegistry implements OnDestroy {
   private _resolvers: IconResolver[] = [];
 
   /**
-   * The CSS class to apply when an `<sbb-icon>` component has no icon name, url, or font specified.
-   * The default is 'sbb-icons'.
+   * The CSS classes to apply when an `<sbb-icon>` component has no icon name, url, or font
+   * specified.
    */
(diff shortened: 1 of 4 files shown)


kyubisation (approved):
LGTM 👍 
Since this is technically a breaking change, should we freeze this, until we stop working on v13?

mhaertwig (commented):
Or we allow both string and array. I find it confusing that ` setDefaultFontSetClass` (singular) accepts an array of strings.

kyubisation:
While that might be confusing, it is not a breaking change (since it allow rest args, e.g. `setDefaultFontSetClass('class-1')` or `setDefaultFontSetClass('class-1', 'class-2')`), however, the return type of `getDefaultFontSetClass` changed from `string` to `string[]`. Or did I misunderstand you?

mhaertwig:
No, you are right. It's only the getter that changed.

jeripeierSBB:
Since the impact is very small and no one uses this (see code.sbb.ch), in my opinion we should merge this. I suggest to freeze as little PRs as possible.

kyubisation:
Agreed 👍

mhaertwig (approved):

PR #1228, opened 2022-03-01
fix(angular/i18n): fix translations not used on production

jeripeierSBB:


src/angular/core/common-behaviors/common.module.ts
```diff
@@ -32,6 +34,18 @@ export interface GranularSanityChecks {
   version: boolean;
 }
 
+/**
```
kyubisation, line 37: Should we maybe move this to after the module to the after functions?

src/angular/core/common-behaviors/common.module.ts
```diff
@@ -75,6 +89,8 @@ export class SbbCommonModule {
         }
       }
     }
+
+    patch$localize();
```
kyubisation, line 93: Should this be moved to the other setup call before the if check?

kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM

PR #1233, opened 2022-03-02
fix(angular/menu): use narrower value for aria-haspopup on trigger e…

mhaertwig:
…lement

Currently we set `aria-haspopup` correctly on the menu trigger, however [according to the spec](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup) we can narrow the value down so that it indicates what kind of popup will be opened.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1236, opened 2022-03-02
fix(angular/datepicker): use aria-live over cdkAriaLive on period button

mhaertwig:
On the period button on the calendar, use `aria-live` over `cdkAriaLive` because that seems to work better with VoiceOver. This fixes an issue where VoiceOver did not announce the month after clicking the "Month" / "Previous Month" buttons


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1237, opened 2022-03-02
fix(angular/dialog): use passed in ComponentFactoryResolver to resolve dialog content

mhaertwig:
Currently the `ComponentFactoryResolver` from the dialog config is only used for the dialog container. These changes switch to also use it for the dialog content component.


jeripeierSBB (approved):
LGTM

PR #1243, opened 2022-03-03
feat(angular/schematics): set lean test configuration on ng-update migration

jeripeierSBB:


src/angular/schematics/ng-add/clean-up.ts
```diff
@@ -7,6 +7,7 @@ import {
 
 import { DuplicateMigration } from './migrations/duplicate-migration';
 
+// noinspection JSUnusedGlobalSymbols
```
kyubisation, line 10: Why is this necessary?

kyubisation (approved):
LGTM 👍

src/angular/schematics/ng-add/clean-up.ts
jeripeierSBB: Intellij warns me about unused function

kyubisation: Ok 👍

PR #1249, opened 2022-03-03
feat(angular/menu): allow grouping of elements

### LINKED ISSUE #201: Feature request: dropdown / contextmenu is missing category/sub-title labels
The usermenu and contextmenu do not yet support the insertion of sub titles / categories - see screenshots:
image
image

Implemented as in design system. Optimally, the labels would allow to contain custom elements.

mhaertwig:
Closes #201

### DIFF
--- a/src/components-examples/angular/menu/menu-grouping/menu-grouping-example.html
+++ b/src/components-examples/angular/menu/menu-grouping/menu-grouping-example.html
@@ -1,10 +1,10 @@
 <button [sbbContextmenuTriggerFor]="menu" aria-label="Show stations" type="button"></button>
 <sbb-menu #menu="sbbMenu">
-  <span tabindex="0" class="sbb-label">Railway stations in Bern</span>
-  <button sbb-menu-item type="button">Bern HB</button>
-  <button sbb-menu-item type="button">Bern Wankdorf</button>
-  <span class="sbb-label">Railway stations in Zurich</span>
-  <button sbb-menu-item type="button">Zurich HB</button>
-  <button sbb-menu-item type="button">Zurich Stadelhofen</button>
-  <button sbb-menu-item type="button">Zurich Oerlikon</button>
+  <span class="sbb-label" id="stations-bern">Railway stations in Bern</span>
+  <button sbb-menu-item type="button" aria-describedby="stations-bern">Bern HB</button>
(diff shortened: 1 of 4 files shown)


src/components-examples/angular/menu/menu-grouping/menu-grouping-example.html
```diff
@@ -0,0 +1,10 @@
+<button [sbbContextmenuTriggerFor]="menu" aria-label="Show stations" type="button"></button>
+<sbb-menu #menu="sbbMenu">
+  <label class="sbb-label">Railway stations in Bern</label>
```
kyubisation, line 3: Maybe just use a span here? I'm uncertain about accessibility implications when using a label.

jeripeierSBB: I'd also suggest no to use label here. maybe in a span or div we should provide a proper aria role. I found `doc-subtitle` but I'm not sure if this would be appropriate.

src/components-examples/angular/menu/menu-grouping/menu-grouping-example.html
```diff
@@ -0,0 +1,10 @@
+<button [sbbContextmenuTriggerFor]="menu" aria-label="Show stations" type="button"></button>
+<sbb-menu #menu="sbbMenu">
+  <span class="sbb-label" id="stations-bern">Railway stations in Bern</span>
+  <button sbb-menu-item type="button" aria-describedby="stations-bern">Bern HB</button>
+  <button sbb-menu-item type="button" aria-describedby="stations-bern">Bern Wankdorf</button>
+  <span class="sbb-label" id="stations-zurich">Railway stations in Zurich</span>
+  <button sbb-menu-item type="button" aria-describedby="stations-zurich">Zurich HB</button>
```
jeripeierSBB, line 7: I think this solution could be ok. what I originally meant, was that you keep your div with the role list, but you set aria-describedby on the list.

jeripeierSBB: Now I understand why the screenreader reads every label again :)

mhaertwig: Okay, sorry. This was a hard one 😄

jeripeierSBB (approved):
LGTM, although in Voiceover the titles won't be read, but I think it could be ok.

PR #1251, opened 2022-03-03
feat(angular/autocomplete): enable hints for option groups inside menu

### LINKED ISSUE #865: Feature request: autocomplete-hint in option-group
When splitting result lists with `sbb-option-group`s it is not possible to give an indication for each group how many items are not displayed

Allow for the autocomplete-hint to be included in option-groups

mhaertwig:
Closes #865

### DIFF
--- a/src/angular/styles/_typography.scss
+++ b/src/angular/styles/_typography.scss
@@ -2499,7 +2499,7 @@ $menuItemPaddingTopBottom: 4;
     }
   }
 
-  .sbb-label {
+  .sbb-label:not(.sbb-autocomplete-hint) {
     padding-top: calc(#{pxToRem(22)} * var(--sbb-scaling-factor));
 
     :where(html.sbb-lean) & {
@@ -2515,7 +2515,7 @@ $menuItemPaddingTopBottom: 4;
       margin-top: pxToRem(4);
     }
 
-    .sbb-label {
+    .sbb-label:not(.sbb-autocomplete-hint) {
       padding-top: calc(#{pxToRem(16)} * var(--sbb-scaling-factor));
 
       :where(html.sbb-lean) & {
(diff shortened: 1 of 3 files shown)


src/angular/styles/_typography.scss
```diff
@@ -2499,7 +2499,7 @@ $menuItemPaddingTopBottom: 4;
     }
   }
 
-  .sbb-label {
+  .sbb-label:not(.sbb-autocomplete-hint) {
```
kyubisation, line 2502: I'm uncertain about using `:not` here, but also not opposed to it. Should maybe the `.sbb-label` styles also be applied to the `.sbb-autocomplete-hint`?
@jeripeierSBB, @mhaertwig Opinions?

jeripeierSBB: sbb-autocomplete-hint has the sbb-label styles applied, so here I think it's to exclude the padding. I'd suggest to just overwrite padding in sbb-autocomplete-hint if possible

kyubisation, line 2502:
```suggestion
  .sbb-label:where(:not(.sbb-autocomplete-hint)) {
```

src/angular/styles/_typography.scss
```diff
@@ -2515,7 +2515,7 @@ $menuItemPaddingTopBottom: 4;
       margin-top: pxToRem(4);
     }
 
-    .sbb-label {
+    .sbb-label:not(.sbb-autocomplete-hint) {
```
kyubisation, line 2518:
```suggestion
    .sbb-label:where(:not(.sbb-autocomplete-hint)) {
```

kyubisation (approved):
LGTM 👍

PR #1256, opened 2022-03-04
fix(angular/select): disabled state out of sync when swapping form group with a disabled one

kyubisation:
Fixes the disabled state of a `sbb-select` falling out of sync with its form control if the
control's group is swapped out with one that is disabled on init.


jeripeierSBB (approved):
LGTM

mhaertwig:
LGTM

mhaertwig (approved):

PR #1261, opened 2022-03-07
fix(angular/button): avoid setting a tabindex on all link buttons

mhaertwig:
Fixes that the button component was always setting a `tabindex="0"` when it's placed on a link element. This is unnecessary, because links are in the tab order by default.

We actually had a test that has an assertion against this, but it was passing by accident because we weren't running change detection before making the asserting.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1262, opened 2022-03-07
fix(angular/dialog): don't block child component animations on open

mhaertwig:
Reworks several components that usually contain projected content not to block animations within that content. This prevents things like expansion panels from appearing as open.

Demo: https://stackblitz.com/edit/angular-7yh3kk?file=src%2Fapp%2Fcomponent-data-dialog-example.html


kyubisation (approved):
LGTM 👍

PR #1264, opened 2022-03-08
fix(angular/tooltip): decouple removal logic from change detection

mhaertwig:


src/angular/tooltip/tooltip.spec.ts
```diff
@@ -585,8 +573,10 @@ describe('SbbTooltip', () => {
       fixture.detectChanges();
       tick(500);
       fixture.detectChanges();
+      finishCurrentTooltipAnimation(overlayContainerElement, false);
 
       expect(tooltipDirective._isTooltipVisible()).toBe(false);
+      console.log(overlayContainerElement, overlayContainerElement.textContent);
```
kyubisation, line 579: Intended?

kyubisation (approved):
LGTM 👍 
This is also technically a breaking change, but I'm not sure whether we would really break a project.
I'm unsure, whether we should already consider our idea of splitting of maintenance branches and auto merges to them?
@mhaertwig, @jeripeierSBB We should probably discuss this tomorrow.

src/angular/tooltip/tooltip.scss
```diff
@@ -52,6 +58,34 @@
...
+    opacity: 1;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
+
+.sbb-tooltip-container-show {
+  animation: sbb-tooltip-container-show 200ms cubic-bezier(0, 0, 0.2, 1) forwards;
```
jeripeierSBB, line 82:
```suggestion
  animation: sbb-tooltip-container-show var(--sbb-transition-duration-fast) cubic-bezier(0, 0, 0.2, 1) forwards;
```

jeripeierSBB (approved):
LGTM, thanks a lot!

jeripeierSBB:
@mhaertwig @kyubisation I still like to postpone branch splits :). For me this would be another case where we could merge it without to expect a problem with an application.

PR #1265, opened 2022-03-08
chore: test if I can write to this repo

swiss-chris:
test if I can write to this repo

why would I want a footer ?


swiss-chris:
Closing again to prevent confusion. context: JIRA/FLOW story ROKAS-481

PR #1266, opened 2022-03-08
fix(angular/autocomplete): correct color of hover elements

mhaertwig:
When hovering over an autocomplete item, the `strong` part does not change its color. This is because the selector `.sbb-menu-item:is(:hover strong)` did not select the `strong` tag.

### DIFF
--- a/src/angular/styles/_typography.scss
+++ b/src/angular/styles/_typography.scss
@@ -2437,12 +2437,13 @@ $menuItemPaddingTopBottom: 4;
     @include user-select(none);
   }
 
-  &:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus, :hover strong, :focus
-        strong), .sbb-focused) {
-    color: var(--sbb-color-red);
+  &:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus), .sbb-focused) {
+    :is(&, & strong) {
+      color: var(--sbb-color-red);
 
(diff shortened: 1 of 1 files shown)


src/angular/styles/_typography.scss
```diff
@@ -2437,12 +2437,13 @@ $menuItemPaddingTopBottom: 4;
     @include user-select(none);
   }
 
-  &:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus, :hover strong, :focus
-        strong), .sbb-focused) {
-    color: var(--sbb-color-red);
+  &:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus), .sbb-focused) {
```
jeripeierSBB, line 2440: The generated output is a little bit long, but I don't see a good alternative...

generated output:
`:is(.sbb-menu-item:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus), .sbb-focused),
      .sbb-menu-item:is(:not(:disabled):not([disabled]):not(.sbb-disabled):is(:hover, :focus), .sbb-focused) strong)`

jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

src/angular/styles/_typography.scss
kyubisation: I agree, but in this case I think we just need to live with it.

PR #1269, opened 2022-03-09
fix(angular/autocomplete): able to tab into descendants with visibility while closed

mhaertwig:
The expansion panel sets `visibility: hidden` while it's closed in order to prevent users from tabbing into the content. This breaks down if a child has its own `visibility`, because it overrides the one coming from the parent. We can't use `display` in the animation definition, because it prevents the animations module from calculating the height when animating.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1273, opened 2022-03-10
fix(angular/chips): create new array or set on update

mhaertwig:
Before this change the chip control modified the underlying array or set instance in place. This led to unexpected behaviour, e.g. when used with `distinctUntilChanged((prev, curr) => prev === curr)` (where `prev` and `curr` point to the same array). Therefore, we now return a new array or set instance after a chip was removed or added.

### DIFF
--- a/src/angular/chips/chip-input.ts
+++ b/src/angular/chips/chip-input.ts
@@ -270,7 +270,7 @@ export class SbbChipInput implements SbbChipTextControl, OnChanges, OnDestroy, A
     const isArray = Array.isArray(currentCollection);
     const isSet = currentCollection instanceof Set;
     if (isArray) {
-      control.patchValue([...currentCollection, inputValue]); // .push(inputValue);
+      control.patchValue([...currentCollection, inputValue]);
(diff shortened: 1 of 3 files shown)


src/angular/chips/chip-input.ts
```diff
@@ -270,12 +270,11 @@ export class SbbChipInput implements SbbChipTextControl, OnChanges, OnDestroy, A
     const isArray = Array.isArray(currentCollection);
     const isSet = currentCollection instanceof Set;
     if (isArray) {
-      currentCollection.push(inputValue);
+      control.patchValue([...currentCollection, inputValue]); // .push(inputValue);
```
jeripeierSBB, line 273:
```suggestion
      control.patchValue([...currentCollection, inputValue]);
```

jeripeierSBB (changes requested):

jeripeierSBB (approved):
LGTM, thanks for the tests

kyubisation (approved):
LGTM 👍 
Nice catch and clean solution 😃

PR #1276, opened 2022-03-10
fix(angular/autocomplete): always emit closed event

### LINKED ISSUE #1274: Bug: Autocomplete opened events inconsistent
**Describe the bug**
The "closed" event of the autocomplete event is missing if - after filtering - there are no options displayed when the focus is lost.

**To Reproduce**
1. Go to https://stackblitz.com/edit/angular-84fpps?file=src/app/autocomplete-reactive-forms-example.html
2. Click enter "asdf" in the input
3. Click somewhere to leave the input field (blur)
4. Observe that no "closed" event is emitted

**Expected behavior**
The "closed" event should be emitted when the user moves away from the input (not when there are no options).

**Screenshots**
autocomplete-close-event

 - OS: win10
 - Browser: chrome
 - Version: SACL 13.8.0

**Additional context**
We need to implement a lazy computation of the options to optimize performance. For this we need consistent opened/closed events to know when we need to compute the current list of options.

mhaertwig:
If the user enters a string that does not match any option, the `closed` event was not emitted before. This PR ensures that a `closed` event is emitted on blur if the panel is still open.

Closes #1274

### DIFF
--- a/src/angular/autocomplete/autocomplete-trigger.ts
+++ b/src/angular/autocomplete/autocomplete-trigger.ts
@@ -111,7 +111,7 @@ export function getSbbAutocompleteMissingPanelError(): Error {
     // Note: we use `focusin`, as opposed to `focus`, in order to open the panel
     // a little earlier. This avoids issues where IE delays the focusing of the input.
     '(focusin)': '_handleFocus()',
-    '(blur)': '_onTouched()',
+    '(blur)': '_handleBlur()',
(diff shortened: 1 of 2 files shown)


jeripeierSBB (changes requested):
As discussed, this change fires unwanted closed events, e.g. on browser tab change or when an option is regularly selected (then fired twice)

PR #1283, opened 2022-03-15
fix(angular/tabs): avoid timeouts in background tabs

mhaertwig:
The tabs have a couple of `requestAnimationFrame` calls when checking for the size/position of elements which can cause tests to hang when they're in a background tab. The problem is that browsers block `requestAnimationFrame` when the browser is out of focus and test harnesses will wait for the call to resolve.

These changes switch to `NgZone.onStable` in an attempt to resolve the issue.

Moreover, remove a duplicate test.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1284, opened 2022-03-15
feat(angular/dialog): add the ability to control the animation duration

mhaertwig:
Since the dialog animation is on the `SbbDialogContainer`, consumers aren't able to disable the animation. These changes add properties to the dialog config that allow consumers to set the duration of the dialog's enter and exit animations.

**Memo**: Merge #1287 first

### DIFF
--- a/src/angular/dialog/dialog-animations.ts
+++ b/src/angular/dialog/dialog-animations.ts
@@ -27,14 +27,17 @@ export const sbbDialogAnimations: {
     transition(
       '* => enter',
       group([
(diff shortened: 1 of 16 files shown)


jeripeierSBB (changes requested):
As the dialog is quite an equivalent to the lightbox, the changes should be adapted there too, respectively implemented in the base class. Additionally an example in the lightbox section would be nice too.

kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1301, opened 2022-03-16
fix(angular/icon): clip overflowing icon elements

jeripeierSBB:
Clips any elements that are overflowing an icon. This avoids issues where parts of the icon could stick out and layouts.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1314, opened 2022-03-21
fix(angular/autocomplete): re-enter the Angular zone when the `NgZone.onStable` emits

jeripeierSBB:
The `NgZone.onStable` always emits outside of the Angular zone, but the zone has not been
re-entered. This leads to change detection being called outside of the Angular zone and the
`autocomplete.opened` also was emitting values outside of the Angular zone.


mhaertwig (approved):
LGTM 👍

kyubisation (approved):
LGTM 👍

PR #1317, opened 2022-03-21
fix(angular/schematics): mark ng-add schematics as hidden

jeripeierSBB:
These schematics are meant to be used with `ng-add` and shouldn't be made visible. By setting these schematics hidden we don't show them in the `ng generate` help output.


kyubisation (approved):
LGTM 👍

PR #1318, opened 2022-03-21
refactor: replace deprecated String.prototype.substr()

jeripeierSBB:
.substr() is deprecated, so we replace it with .slice() which works similarly but isn't deprecated


mhaertwig (approved):
LGTM 👍

PR #1323, opened 2022-03-22
fix(angular/checkbox): model value not updated when using toggle method

mhaertwig:
The checkbox doesn't update its `ControlValueAccessor` value when it is toggled via the `toggle` method.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1324, opened 2022-03-22
fix(angular/select): value set through property not being propagated to value accessor

mhaertwig:
Fixes values set through the `value` property not being propagated to the `value` in the `ControlValueAccessor`.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1332, opened 2022-03-24
refactor(angular/form-field): allow AbstractControlDirective as MatFormFieldControl.ngControl

mhaertwig:
Allows for an `AbstractControlDirective` to be passed in as the `SbbFormFieldControl.ngControl`. The advantage of passing in the abstract control directive is that it allows both form controls and form groups. This is necessary for the date range picker.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1343, opened 2022-03-28
fix(angular/tabs): focus wrapping back to selected label when using shift + tab

jeripeierSBB:
Currently, the `tabindex` of each of the tab labels is determined by the selected index. This means that if the user tabbed into the header, pressed the right arrow and then pressed shift + tab, their focus would end up on the selected tab. These changes switch to basing the `tabindex` on the focused index which is tied both to the selected index and the user's keyboard navigation.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1344, opened 2022-03-28
fix(angular/tabs): update tab state when active tab is swapped out

mhaertwig:
Fixes that when the selected tab is swapped out, we weren't marking the new tab as active and emitting the `selectedTabChange` event.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1345, opened 2022-03-28
fix(angular/radio): set tabindex based on selected state

mhaertwig:
Currently all radio buttons inside a radio group have a `tabindex` of 0,  unless they're disabled, and we leave it up to the browser to focus the proper button based on its selected state when the user is tabbing. This works for the most part, but it breaks down with something like our focus trap which determines which element to focus based on its `tabindex` since all buttons have the same `tabindex`. These changes fix the issue by setting a `tabindex` of 0 only on the selected radio button.

### DIFF
--- a/src/angular/radio-button/radio-button.spec.ts
+++ b/src/angular/radio-button/radio-button.spec.ts
@@ -60,6 +60,17 @@ class StandaloneRadioButtons {
 
 @Component({
   template: `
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1347, opened 2022-03-29
ci: implement maintenance workflow

mhaertwig:
This pull request uses GitHub actions to
- check for `@breaking-change` annotations
- add the "target: maintenance" label if the pr is not a breaking change
- cherry-picks non-breaking changes into the maintenance branch
- collects failed cherry-picks in an issue

### DIFF
--- a/.github/workflows/maintenance-cherry-pick-workflow.yml
+++ b/.github/workflows/maintenance-cherry-pick-workflow.yml
@@ -1,0 +1,40 @@
+name: cherry-pick on maintenance branch
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  cherry_pick:
+    runs-on: ubuntu-latest
+    outputs:
+      output: ${{ steps.cherry_pick.outputs.output }}
+    if: |
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'target: maintenance')
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Cherry-pick changes into maintenance"
+      id: cherry_pick
+      run: |
+        git checkout maintenance
+        echo "::set-output name=output::$(git -c user.name="cherry.picker" \
+          -c user.email="cherry.picker@sbb.ch" cherry-pick ${{ github.sha }} | xargs)"
+        git push
+  update_maintenance_issue:
+    runs-on: ubuntu-latest
+    needs: cherry_pick
+    if: ${{ contains(needs.cherry_pick.outputs.output, 'CONFLICT') }}
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+    - uses: ./.github/actions/yarn-install
+    - name: "Update maintenance issue"
+      run: yarn ts-node scripts/update-maintenance-issue.ts
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        PR_NUMBER: ${{ github.event.number }}
--- a/.github/workflows/maintenance-tagging-workflow.yml
+++ b/.github/workflows/maintenance-tagging-workflow.yml
@@ -1,0 +1,26 @@
+name: Add or remove maintenance labels
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Get @breaking-change annotation counts"
+      id: deprecations
+      run: echo "::set-output name=count::$(git diff ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | egrep "^\+" | grep '@breaking-change' -c)"
+    - name: "Add maintenance label"
+      uses: actions-ecosystem/action-add-labels@v1
+      if: steps.deprecations.outputs.count == '0'
+      with:
+        labels: "target: maintenance"
+    - name: "Remove maintenance label"
+      uses: actions-ecosystem/action-remove-labels@v1
+      if: steps.deprecations.outputs.count != '0'
+      with:
+        labels: "target: maintenance"
--- a/scripts/update-maintenance-issue.ts
+++ b/scripts/update-maintenance-issue.ts
@@ -1,0 +1,61 @@
+import { Octokit } from 'octokit';
+
+const githubToken = process.env['GITHUB_TOKEN'];
+const pullRequestNumber = parseInt(process.env['PR_NUMBER']!, 10);
+
+const repoConfig = {
+  owner: 'sbb-design-systems',
+  repo: 'sbb-angular',
+};
+
+const issuePath = {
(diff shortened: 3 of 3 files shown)


.github/workflows/maintenance-cherry-pick-workflow.yml
```diff
@@ -0,0 +1,40 @@
...
+      contains(github.event.pull_request.labels.*.name, 'target: maintenance')
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Cherry-pick changes into maintenance"
+      id: cherry_pick
+      run: |
+        git checkout maintenance
+        echo "::set-output name=output::$(git -c user.name="cherry.picker" \
```
jeripeierSBB, line 23: is it not possible to remain the original committer?

scripts/update-maintenance-issue.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    if (!pr.data.title || !pr.data.html_url) {
+      throw new Error('Could not load pull request');
+    }
+
+    const hint = '**Unable to merge the following pull requests into the maintenance branch**';
+    const dateInfo = `${this._now.toISOString()}`;
+    let openTasks = this._extractOpenTasks(issue.data.body);
+    openTasks.push(`- [ ] [${pr.data.title}](${pr.data.html_url})`);
+    openTasks = [...new Set(openTasks)];
```
jeripeierSBB, line 40: maybe extract method as it is in extract, but this time the opposite (include duplication removal and line joining?

scripts/update-maintenance-issue.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return this._octokit.rest.issues.update({
+      ...issuePath,
+      body: `${hint}\r\n${openTasks.join('\r\n')}\r\n${dateInfo}`,
+    });
+  }
+
+  private _extractOpenTasks(issueBody: string = ''): string[] {
+    return issueBody.split('\r\n').filter((line) => line.startsWith('- [ ]'));
```
jeripeierSBB, line 49: Is `\n` not sufficient?

.github/workflows/maintenance-tagging-workflow.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Get @breaking-change annotation counts"
+      id: deprecations
```
jeripeierSBB, line 15:
```suggestion
      id: breaking-changes
```

.github/workflows/maintenance-tagging-workflow.yml
```diff
@@ -0,0 +1,26 @@
...
+      run: echo "::set-output name=count::$(git diff ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | egrep "^\+" | grep '@breaking-change' -c)"
```
jeripeierSBB, line 16: I thought we only want to check for commit messages, if there is written "BREAKING CHANGE:" ? @kyubisation What's your opinion?

.github/workflows/maintenance-cherry-pick-workflow.yml
mhaertwig: image
That's how it's displayed. We can see that it was cherry-picked by the bot.

.github/workflows/maintenance-tagging-workflow.yml
kyubisation: Yes, just checking the PR message should work best.
To explain: We sometimes add a note `@breaking-change 15` to indicate that something should be changed for version 15, which will be a breaking change, but is not with the current change.

.github/workflows/maintenance-cherry-pick-workflow.yml
```diff
@@ -0,0 +1,40 @@
...
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'target: maintenance')
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Cherry-pick changes into maintenance"
+      id: cherry_pick
+      run: |
+        git checkout maintenance
```
kyubisation, line 22: Maybe this is more a general question about maintenance branch naming.
I'd prefer for each (current and new) major version to create a dedicated branch (e.g. `13.x` and `14.x`) and to update the target branch when creating the first major release.
@jeripeierSBB, @mhaertwig Thoughts?

.github/workflows/maintenance-cherry-pick-workflow.yml
```diff
@@ -0,0 +1,40 @@
...
+        echo "::set-output name=output::$(git -c user.name="cherry.picker" \
+          -c user.email="cherry.picker@sbb.ch" cherry-pick ${{ github.sha }} | xargs)"
```
kyubisation, line 24: I do wonder whether it is wise to use a mail address that does not exist.
Maybe just use `user.name="github-actions" user.email="github-actions@github.com"`

See https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

scripts/update-maintenance-issue.ts
```diff
@@ -0,0 +1,61 @@
...
+  ...repoConfig,
+  issue_number: 1346,
+};
+
+const prPath = {
+  ...repoConfig,
+  pull_number: pullRequestNumber,
+};
+
+export class MaintenanceIssueUpdater {
```
kyubisation, line 21: nitpick: the class does not need to be exported

kyubisation (changes requested):
Nice work so far 👍

.github/workflows/maintenance-cherry-pick-workflow.yml
kyubisation: Should we maybe have corresponding labels so it would be transparent where it would be cherry-picked to?
e.g. `target: 13.x` cherry-picks to `13.x` and `target: 14.x` cherry-picks to `14.x` and if both labels are set, it cherry-picks to both? The auto-label would only assign the manually configured label and we can switch it, once we want to move to the next major version.
But I'm open to discussion, if this seems to complex.

.github/workflows/maintenance-tagging-workflow.yml
```diff
@@ -0,0 +1,29 @@
+name: Add or remove maintenance labels
+
+on:
+  pull_request:
+    branches: [master]
```
mhaertwig, line 5:
```suggestion
    branches: [master] # target branch
```

.github/workflows/maintenance-tagging-workflow.yml
```diff
@@ -0,0 +1,24 @@
...
+
+env:
+  TARGET_RELEASE: 13.x
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Add target label"
+        if: ${{ contains(github.event.pull_request.body, 'BREAKING CHANGE:') }}=
```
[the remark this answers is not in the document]
mhaertwig: Of course, thanks!

kyubisation (approved):
LGTM 👍

PR #1360, opened 2022-04-01
fix(angular/tabs): wrong scroll distance if selected tab is removed

kyubisation:
Fixes the case where the user has scrolled to the end of the tab list and the selected tab from the beginning is removed. Since we no longer have a selected tab, the scroll distance wasn't being updated.


jeripeierSBB (approved):
LGTM

mhaertwig (approved):
LGTM 👍

PR #1361, opened 2022-04-01
fix(angular/datepicker): avoid rerender when min/maxDate changes to different time on the same day

kyubisation:
Avoid re-rendering when the [minDate] or [maxDate]
Input change to a different time on the same day.

In ngOnChanges, do not call init if the previous and current value for
minDate/maxDate are on the same day.

### DIFF
--- a/src/angular/datepicker/calendar/calendar.ts
+++ b/src/angular/datepicker/calendar/calendar.ts
@@ -281,7 +281,7 @@ export class SbbCalendar<D> implements AfterContentInit, AfterViewChecked, OnDes
   ngOnChanges(changes: SimpleChanges) {
     // Ignore date changes that are at a different time on the same day. This fixes issues where
     // the calendar re-renders when there is no meaningful change to [minDate] or [maxDate]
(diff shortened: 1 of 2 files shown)


src/angular/datepicker/calendar/calendar.ts
```diff
@@ -278,7 +279,21 @@ export class SbbCalendar<D> implements AfterContentInit, AfterViewChecked, OnDes
   }
 
   ngOnChanges(changes: SimpleChanges) {
-    const change = changes.minDate || changes.maxDate || changes.dateFilter;
+    // Ignore date changes that are at a different time on the same day. This fixes issues where
+    // the calendar re-renders when there is no meaningful change to [minDate] or [maxDate]
+    // (#24435).
```
jeripeierSBB, line 284:
```suggestion
    // (angular/components#24435).
```

jeripeierSBB (approved):
LGTM

mhaertwig (approved):
LGTM 👍

PR #1371, opened 2022-04-05
refactor(angular/dialog): use embedded injector to provide ref to template dialogs

jeripeierSBB:
Previously we had to walk the DOM in order to figure out which dialog ref belonged to a specific template dialog. We no longer need to do that if we provide an injector to the template portal.

### DIFF
--- a/src/angular/dialog/dialog.ts
+++ b/src/angular/dialog/dialog.ts
@@ -271,6 +271,7 @@ export abstract class _SbbDialogBase<
           <any>{
             $implicit: config.data,
             dialogRef,
+            lightboxRef: dialogRef,
           },
           injector
         )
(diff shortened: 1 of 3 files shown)


src/angular/dialog/dialog.ts
```diff
@@ -261,17 +261,21 @@ export abstract class _SbbDialogBase<
...
-          $implicit: config.data,
-          dialogRef,
-          lightboxRef: dialogRef,
-        })
+        new TemplatePortal<T>(
+          componentOrTemplateRef,
+          null!,
+          <any>{
+            $implicit: config.data,
+            dialogRef,
```
kyubisation, line 273: I think this change broke the lightbox test.
Attempted fix:
```suggestion
            dialogRef,
            lightboxRef: dialogRef,
```

kyubisation (changes requested):

kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1372, opened 2022-04-05
refactor(angular/core): no longer define stateChanges in `mixinErrorState`

jeripeierSBB:
The error state mixin provided in `@sbb-esta/angular/core` currently
defines the `stateChanges` class member as part of the mixin. This
is unexpected as consumers might deal with the `stateChanges` member
differently. e.g. in some components the `stateChanges` field is
intended to show up in the docs, or the JSDoc description varies. e.g.
the observable could emit whenever form-field state changes, and it
should be updated, or it emits always when something changes (e.g.
even a component input which is not relevant for the form-field
control).

In general, we want to avoid this member being defined in the mixin
as the mixin is rather about the error state, and not defining a
subject that can emit whenever "state" changes.

BREAKING CHANGE: Previously the `mixinErrorState` mixin function defined
a class member for `stateChanges`. This is no longer the case, and consumers
need to provide the `stateChanges` class member themselves.

### DIFF
--- a/src/angular/chips/chip-list.ts
+++ b/src/angular/chips/chip-list.ts
@@ -33,11 +33,22 @@ import { SBB_CHIP, SBB_CHIP_LIST } from './chip-tokens';
 // tslint:disable-next-line:naming-convention
 const _SbbChipListBase = mixinErrorState(
   class {
+    /**
+     * Emits whenever the component state changes and should cause the parent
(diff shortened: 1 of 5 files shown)


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1381, opened 2022-04-08
feat(journey-maps): migrate web component

sebisteiner:


src/journey-maps/web-component/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+)
+
+esbuild(
+    name = "bundle",
+    args = {
+        "resolveExtensions": [
+            ".mjs",
+            ".js",
+        ],
+        "sourcemap": True,
```
sebisteiner, line 30: @kyubisation Wenn ich `"sourcemap": False` setze, schlägt der Build fehl, weil keinen Source Map erstellt wird. Es läuft nur durch, wenn ich gleichzeitig auch `output_dir = True` setze. Aber das möchte ich eigentlich nicht.

Hast Du eine Idee, woran das liegt?

src/journey-maps/angular/components/template-renderer/template-renderer.ts
```diff
@@ -0,0 +1,68 @@
...
+  Input,
+  OnChanges,
+  SimpleChanges,
+  TemplateRef,
+  ViewChild,
+} from '@angular/core';
+
+import { SbbTemplateType } from '../../journey-maps.interfaces';
+
+@Component({
```
kyubisation, line 14: At first glance a structural directive might be a better match for this functionality.
(e.g. maybe extend https://github.com/angular/angular/blob/13.3.3/packages/common/src/directives/ng_template_outlet.ts#L10-L71 )
Or is there a specific reason why you prefer a wrapper component?

src/journey-maps/angular/components/template-renderer/template-renderer.ts
```diff
@@ -0,0 +1,68 @@
...
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!changes.template) {
+      return;
+    }
+
+    this._htmlTemplateObserver?.disconnect();
+
+    if (typeof this.template === 'string') {
+      this.ngTemplate = undefined;
+      const content = (document.getElementById(this.template) as HTMLTemplateElement).content;
```
kyubisation, line 45: This might not work in a project that (extensively) uses Shadow DOM, as it does not pierce Shadow DOM.
But I don't have an alternative suggestion at the moment.

src/journey-maps/angular/components/template-renderer/template-renderer.ts
```diff
@@ -0,0 +1,68 @@
...
+      const setupHtmlTemplate = () => {
+        this.htmlTemplate = content.cloneNode(true) as HTMLElement;
+        this.addHtmlTemplateToDOM();
+      };
+
+      setupHtmlTemplate();
+      this._htmlTemplateObserver = new MutationObserver(setupHtmlTemplate);
+      this._htmlTemplateObserver.observe(content, {
```
kyubisation, line 53: I believe this would not trigger for changes in attributes or text.
The following should work but is also more expensive.
```ts
const contentObserverConfig: MutationObserverInit = {
  attributes: true,
  childList: true,
  subtree: true,
  characterData: true,
};
```

src/journey-maps/angular/components/template-renderer/template-renderer.ts
sebisteiner: > Or is there a specific reason why you prefer a wrapper component?
No - done.

src/journey-maps/angular/components/template-renderer/template-renderer.ts
sebisteiner: Thanks for the input.

Can we leave it like that for the moment and address the issue later?

src/journey-maps/angular/components/template-renderer/template-renderer.ts
sebisteiner: Thanks - I've adopted your suggestion. The templates should not be that big so I think it's ok.

src/journey-maps/angular/components/template-renderer/template-renderer.ts
kyubisation: That's fine. 👍

kyubisation (changes requested):
One minor detail.
Other than that looks good to me. 👍

sebisteiner:
Will reopen a new pull request - had some issues during rebase.

PR #1386, opened 2022-04-12
fix(angular/autocomplete): prevent clean-up from repeatedly being called

kyubisation:
Previously we called clean-up (and option highlighting) on each zone.js
onStable emit. This caused a near constant loop of the functionality
being called on large applications. We fixed this by limiting the call
to just the next onStable emit and removing it entirely for the clean-up
as it is not necessary there.


jeripeierSBB (approved):
LGTM

PR #1393, opened 2022-04-13
fix(angular/chips): prevent default behavior on remove button

jeripeierSBB:
Adds a `preventDefault` call to the `click` handlers of the chip remove icon in order to avoid things like form submissions and link navigations.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1397, opened 2022-04-14
build: use ngssc for variable insertion in showcase image

kyubisation:


jeripeierSBB (changes requested):
I think we should add angular-server-side-configuration as dependency and configure renovate to run ng-update migration

jeripeierSBB (approved):
LGTM, thanks

PR #1409, opened 2022-04-25
fix(angular/autocomplete): outside click in Angular zone

jeripeierSBB:
Fixes a bug in `autocomplete` when outside click doesn't trigger `changeDetection`.


mhaertwig (approved):
LGTM 👍

PR #1434, opened 2022-05-04
refactor(angular/loading): rename loading to loading-indicator

mhaertwig:
This PR renames the loading module and all its symbols to loading-indicator. It is however re-exported from the previous loading entrypoint. It also provides a simple symbol and selector renaming migration. The module name migration does not work reliably and is delayed to version 15.


src/angular/schematics/ng-update/index.ts
```diff
@@ -11,32 +14,34 @@ import { ClassNamesMigration } from '../ng-add/migrations/class-names';
...
-      sbbAngularMigrations,
-      sbbAngularUpgradeData,
-      onMigrationComplete
-    ),
+    createMigrationSchematicRule(TargetVersion.V13, [], sbbAngularUpgradeData, onMigrationComplete),
   ]);
 }
 
 /** Entry point for the migration schematics with target of Angular CDK 14.0.0 */
 export function updateToV14(): Rule {
```
mhaertwig, line 27: FYI: I didn't remove this code because we may need it for renaming the `sbb-autocomplete-hint` to `sbb-option-hint`.

jeripeierSBB (approved):
LGTM

PR #1442, opened 2022-05-05
feat(journey-maps): synchronize click-overlay and selection state

sebisteiner:


jeripeierSBB (approved):
LGTM

src/journey-maps/angular/services/map/events/map-selection-event.ts
```diff
@@ -52,9 +51,9 @@ export class SbbMapSelectionEvent {
     this._subscription?.unsubscribe();
   }
 
-  toggleSelection(eventData: SbbFeaturesClickEventData): void {
```
kyubisation, line 55: Be aware that this is technically a breaking change. Try to avoid breaking changes or document them in the PR message (which requires a major version release).

kyubisation (approved):
This seems more like a fix to me. What exactly is the new feature introduced?

sebisteiner:
> This seems more like a fix to me. What exactly is the new feature introduced?

Agree. I renamed the PR title and commit message.

src/journey-maps/angular/services/map/events/map-selection-event.ts
sebisteiner: This service - and also all of our other services - is an internal service and mustn't be used by our users. Therefore I consider this an internal refactoring and no breaking change. The `@Output` event which will be received by the user has not changed.

Do you agree with that?

kyubisation: Hm, we lack a clear distinction between public and private API for these internal services.
It's fine for now.

PR #1447, opened 2022-05-09
fix(angular/datepicker): add ability to have numeric zero value in input

jeripeierSBB:
Shows valid formatted value for 'falsy' values (exp 0)
Custom DateAdapter (exp TimestampDateAdapter) can have valid numeric 0 value (for timestamp it is 1/1/1970), but datepicker input doesn't show formatted value


kyubisation (approved):
LGTM 👍

mhaertwig (changes requested):
Numeric values are converted to a date.

PR #1448, opened 2022-05-09
fix(angular/menu): focus the first item when opening menu on iOS VoiceOver

jeripeierSBB:
When opening the menu using the iOS VoiceOver screen reader, focus the first item in the menu.
Previously, the first menu item would focus on other screen readers like desktop VoiceOver but
not with iOS VoiceOver.

Waiting until `onStable` seems to fix this.


kyubisation (approved):
LGTM 👍

mhaertwig (approved):
LGTM 👍

PR #1450, opened 2022-05-09
fix(journey-maps): show correct start level for indoor routing

sebisteiner:


mstankala:
Looks OK!

src/journey-maps/angular/services/map/map-leit-poi-service.ts
```diff
@@ -10,9 +10,7 @@ import { SbbMapLeitPoi } from '../../model/map-leit-poi';
 import { SbbMapLeitPoiCreator } from './map-leit-poi-creator';
 import { SBB_EMPTY_FEATURE_COLLECTION } from './map-service';
 
-@Injectable({
-  providedIn: 'root',
-})
+@Injectable()
```
kyubisation, line 13: Is there a reason against using `providedIn: 'root'`?

PR #1462, opened 2022-05-17
refactor: adapt sass to @use instead of @import

kyubisation:
BREAKING CHANGE: Previously our sass code could be used via `@import '@sbb-esta/angular/styles.scss';`. This is no longer possible as we refactored our sass code to switch from `@import` to `@use` (see https://sass-lang.com/documentation/at-rules/use). We also adapted the recommended way to export sass code from a package via an `_index.scss` file in the root of the package. You should now be able use the following code to import our sass code: `@use '@sbb-esta/angular' as sbb;`


mhaertwig (approved):
LGTM 🥇

PR #1463, opened 2022-05-17
feat(angular/dialog): switch to CDK dialog

mhaertwig:
Switches the angular dialog to be based on the CDK dialog.

### DIFF
--- a/src/angular/dialog/dialog-config.ts
+++ b/src/angular/dialog/dialog-config.ts
@@ -100,7 +100,7 @@ export class SbbDialogConfig<D = any> {
   restoreFocus?: boolean = true;
 
   /** Whether to wait for the opening animation to finish before trapping focus. */
-  delayFocusTrap?: boolean = true;
+  delayFocusTrap?: boolean = false;
 
   /** Scroll strategy to be used for the dialog. */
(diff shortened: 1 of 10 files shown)


src/angular/dialog/dialog-config.ts
```diff
@@ -100,7 +100,7 @@ export class SbbDialogConfig<D = any> {
   restoreFocus?: boolean = true;
 
   /** Whether to wait for the opening animation to finish before trapping focus. */
-  delayFocusTrap?: boolean = true;
+  delayFocusTrap?: boolean = false;
```
mhaertwig, line 103: This option avoids that multiple dialogs are opened when clicking enter on the focussed button multiple times. ☕

Due to [this line](https://github.com/sbb-design-systems/sbb-angular/pull/1463/files#diff-94901f34e901d382979babed6bb2dd96f702e26084c19519240a0df8bf2d0340L116) this config option had no effect before. To keep the current behaviour, I changed this default option. This was also necessary to make the tests working.

kyubisation (approved):
LGTM 👍

kyubisation:
Nice work! 😃

jeripeierSBB (approved):
LGTM! Thanks a lot for this work

PR #1466, opened 2022-05-18
build: implement configuration for legacy versions

kyubisation:


jeripeierSBB (commented):
I think we should add a note to our major release check list that we don't forget to set this variable.

mhaertwig (approved):
LGTM 👍 
(besides Jeri's comment that a hint to update the `legacyVersions` variable in `values.yaml` would be nice)

kyubisation:
I added the paramter in our ArgoCD repo.

PR #1480, opened 2022-05-31
refactor(angular/autocomplete-hint): rename autocomplete-hint to option-hint

mhaertwig:
This PR renames the autocomplete-hint to option-hint and moves it to the `SbbOptionModule`. It is however re-exported from the previous entrypoint since the module name migration does not work reliable. The automatic migration is delayed to version 15.


src/angular/core/option/option.module.ts
```diff
@@ -1,5 +1,6 @@
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
+import { SbbOptionHint } from '@sbb-esta/angular/core/option/option-hint';
```
jeripeierSBB, line 3: Here you would have to use a relative path
```suggestion
import { SbbOptionHint } from './option-hint';
```
and re format (otherwise `yarn baz` would generate a wrong entry)

src/angular/autocomplete/public-api.ts
```diff
@@ -1,5 +1,9 @@
 export * from './autocomplete.module';
 export * from './autocomplete';
-export * from './autocomplete-hint';
 export * from './autocomplete-origin';
 export * from './autocomplete-trigger';
+
+export {
+  /** @deprecated Use SbbOptionHint instead */
+  SbbOptionHint as SbbAutocompleteHint,
+} from '../core/option/option-hint';
```
jeripeierSBB, line 9: discuss with lukas if absolute path should be used.

jeripeierSBB (changes requested):

src/angular/autocomplete/public-api.ts
```diff
@@ -1,5 +1,9 @@
 export * from './autocomplete.module';
 export * from './autocomplete';
-export * from './autocomplete-hint';
 export * from './autocomplete-origin';
 export * from './autocomplete-trigger';
+
+export {
+  /** @deprecated Use SbbOptionHint instead */
+  SbbOptionHint as SbbAutocompleteHint,
+} from '@sbb-esta/angular/core/option/option-hint';
```
kyubisation, line 9:
```suggestion
} from '@sbb-esta/angular/core';
```

src/angular/core/option/optgroup.html
```diff
@@ -1,4 +1,4 @@
 <label class="sbb-label sbb-optgroup-label" [id]="_labelId"
   >{{ label }} <ng-content></ng-content
 ></label>
-<ng-content select="sbb-option, ng-container, sbb-autocomplete-hint"></ng-content>
```
kyubisation, line 4: sbb-autocomplete-hint should still work for backwards compatibility

src/angular/core/option/option-hint.ts
```diff
@@ -0,0 +1,21 @@
+import { Component } from '@angular/core';
+
+/**
+ * Option-hint IDs need to be unique across components, so this counter exists outside of
+ * the component definition.
+ */
+let nextId = 0;
+
+@Component({
```
kyubisation, line 9: I think we can convert this to a directive.

src/angular/autocomplete/autocomplete.module.ts
```diff
@@ -6,7 +6,6 @@ import { SbbOptionModule } from '@sbb-esta/angular/core';
 import { SbbCommonModule } from '@sbb-esta/angular/core';
```
kyubisation, line 6: I just noticed a previous error of mine; The imports of SbbCommonModule and SbbOptionsModule are both from `@sbb-esta/angular/core` and could be merged.

kyubisation (changes requested):
Thanks for the work!
I noticed a few things.

src/angular/core/option/optgroup.html
jeripeierSBB: I was saying the same, but it works as it is!

mhaertwig: Yes, it's not needed. However, I don't really understand why and therefore added the selector.

jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1493, opened 2022-06-07
chore: remove angular-t9n from renovate ng update

jeripeierSBB:
angular-t9n does not provide update schematics and therefore `ng update` fails.


kyubisation (approved):
LGTM 👍

PR #1510, opened 2022-06-09
refactor(journey-maps): update MapLibre to V2

sebisteiner:


src/journey-maps/angular/types.d.ts
```diff
@@ -0,0 +1,46 @@
+// TOOD ses: Why doesn't it work with https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/mapbox__point-geometry
+
+declare module '@mapbox/point-geometry' {
```
sebisteiner, line 3: @kyubisation I couldn't get it to work with the definitely typed types. (_Could not find a declaration file for module '@mapbox/point-geometry'_)

Do you have any idea or should I keep it like this for the moment?

sebisteiner:
@kyubisation Can you help me with the failing build? I don't understand the problem yet...

src/journey-maps/angular/components/level-switch/services/map-layer-filter.ts
```diff
@@ -31,20 +31,23 @@ export class SbbMapLayerFilter {
     this._setLayerVisibility('rokas_background_mask', level < 0);
   }
 
-  private _calculateLayerFilter(oldFilter: any[], level: number): any[] {
+  private _calculateLayerFilter(
+    oldFilter: FilterSpecification | void,
```
kyubisation, line 35: void should only be used as a return value
Use undefined, if the value might be undefined

kyubisation (changes requested):
One detail, other than that LGTM
After researching a bit more, I would recommend adding the peer dependencies and just have warnings for the web components consumers. What do you think?

sebisteiner:
> One detail, other than that LGTM After researching a bit more, I would recommend adding the peer dependencies and just have warnings for the web components consumers. What do you think?

Ok for me.

src/journey-maps/package.json
```diff
@@ -13,6 +13,12 @@
     "url": "https://github.com/sbb-design-systems/sbb-angular/issues"
   },
   "homepage": "https://angular.app.sbb.ch",
+  "peerDependencies": {
+    "@angular/common": "0.0.0-NG",
+    "@angular/core": "0.0.0-NG",
+    "@types/geojson": "7946",
+    "maplibre-gl": "2.1"
```
kyubisation, line 20: Should we maybe use a range? Or would you prefer more precise versions?
```suggestion
    "@types/geojson": ">=7946",
    "maplibre-gl": "^2.1.0"
```

kyubisation (approved):
LGTM 👍

src/journey-maps/package.json
sebisteiner: 👍

PR #1546, opened 2022-06-23
fix(multiple): style changes for better accessibility

mhaertwig:
* accordion: corret color of header-indicator icon
* chips: use cursor pointer for draggable chips
* notification: change color of warn notification in lean mode
* status: change color and icon of warn state and icon of invalid state
* time-input: center text for disabled input in standard mode


src/angular/styles/_typography.scss
```diff
@@ -1764,6 +1764,10 @@ textarea {
     line-height: sbb.pxToRem(20);
   }
 
+  .sbb-checkbox & {
```
kyubisation, line 1767: @mcilurzo Are the text colors for checkbox and radio button intentionally different?

kyubisation (changes requested):

jeripeierSBB:
Additionally, I had im my mind, that every usage of color grey should be replaced by granite ?

src/angular/styles/_typography.scss
mhaertwig: I've started a discussion in our Teams channel and temporary removed this code, Later I'll make a new PR for the checkbox and radio button changes.

kyubisation (approved):
LGTM 👍 
Thank you very much!

src/angular/accordion/expansion-panel.scss
```diff
@@ -83,7 +83,7 @@
   :where(html:not(.sbb-lean)) & {
     border: var(--sbb-border-width-thin) solid var(--sbb-color-storm);
     border-radius: 50%;
-    color: var(--sbb-color-granite);
+    color: var(--sbb-color-black);
```
jeripeierSBB, line 86: I see that you re-added it, but I think then we should apply it for lean also? Or should it be placed on a parent container?

jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1552, opened 2022-06-27
fix(angular/processflow): accept AbstractControl in isErrorState

mhaertwig:
Accept AbstractControl instead of FormControl to avoid compiler errors when strict typing is enabled


jeripeierSBB (approved):
LGTM

PR #1570, opened 2022-07-06
fix: remove global $localize usage

kyubisation:
Global $localize usage would be called before our $localize monkey
patch. Inlining the $localize call causes a slight duplication, but this
should be acceptable at this scale.

### DIFF
--- a/src/angular/lightbox/lightbox-content-directives.ts
+++ b/src/angular/lightbox/lightbox-content-directives.ts
@@ -22,8 +22,6 @@ import { SbbLightboxRef } from './lightbox-ref';
 /** Counter used to generate unique IDs for dialog elements. */
 let dialogElementUid = 0;
 
-const closeAriaLabel = $localize`:Aria label to close a dialog@@sbbLightboxCloseLightbox:Close lightbox`;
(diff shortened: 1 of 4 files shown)


src/angular/lightbox/lightbox-content-directives.ts
```diff
@@ -98,7 +97,8 @@ export class SbbLightboxTitle extends _SbbDialogTitleBase implements OnInit {
   @ViewChild(SbbLightboxClose, { static: true }) _lightBoxClose: SbbLightboxClose;
 
   /** Arial label for the close button. */
-  @Input() override closeAriaLabel: string = closeAriaLabel;
+  @Input()
+  override closeAriaLabel: string = $localize`:Aria label to close a dialog@@sbbLightboxCloseLightbox:Close lightbox`;
```
jeripeierSBB, line 101: Should we not add a comment that this string is a duplicate?

kyubisation: Discussed via Teams; Angular compiler would complain if the text was different for the same id.

PR #1574, opened 2022-07-07
fix(tooltip): allow click event bubbling on hover trigger

### LINKED ISSUE #1568: Bug: Hover Tooltip prevents click on trigger element to propagate
**Describe the bug**
Before, when using the tooltip with trigger=hover clicks were propagated to parent elements.
This is neccessary for us, as we include hover tooltips in table cells and have a click-handler on the row level.

**To Reproduce**
1. generate a tooltip with trigger=hover inside a wrapper that has a click handler
2. click on the tooltip trigger
3. observe that the click handler is _not_ invoced

**Expected behavior**
Click handler should be invoked.

**Additional context**
Maybe related to https://github.com/sbb-design-systems/sbb-angular/commit/f84c82a1e293e3099b4e9ff3029d14e840ffc1b8

kyubisation:
Closes #1568

### DIFF
--- a/src/angular/tooltip/tooltip.spec.ts
+++ b/src/angular/tooltip/tooltip.spec.ts
@@ -1264,6 +1264,32 @@ describe('SbbTooltip', () => {
 
       expect(outerClickSpy).not.toHaveBeenCalled();
     }));
+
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM

PR #1580, opened 2022-07-11
docs: update steps for v13 migration

mhaertwig:
Due to https://github.com/angular/angular-cli/issues/22654 the migration from version 12 to version 13 does not work properly. To circumvent the bug `@angular-devkit/build-angular@13` is updated separately with a fixed version.


kyubisation (approved):
LGTM 👍

PR #1585, opened 2022-07-12
fix(angular/search): close and open header search programmatically

### LINKED ISSUE #1579: SbbHeaderSearch: Overlay is not programmatically closeable
### Overview

**Observation**: Invoking the `SbbHeaderSearch.close()` method has no effect:
* There is still an invisible overlay displayed on the whole page content, which has to be clicked on (by the user or programmatically) to get it closed.
* The header search field is still expanded (displaying the input text field and the red search button).

**Expected behavior**:
* There should be no invisible overlay be displayed on the whole page, such that the next click has immediate effect.
* The header search field should collapse back to be displayed as a button without a text field.

### Steps to reproduce

app.component.ts:
```javascript
@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.scss'],
})
export class AppComponent {

  @ViewChild(SbbHeaderSearch) headerSearch!: SbbHeaderSearch;

  searchControl = new FormControl('');

  search(searchExpression: string): void {

    this.searchControl.reset('');
    this.headerSearch.close();
  }
}
```

app.component.html:
```html
<sbb-header-lean label="App title" subtitle="Version 0.1" collapseBreakpoint="mobile">
  <sbb-header-environment>dev</sbb-header-environment>

  <button type="button" sbbHeaderSearch>
    <sbb-search (search)="search($event)" aria-label="Start search">
      <input sbbInput placeholder="Search" [formControl]="searchControl" />
    </sbb-search>
  </button>
  
</sbb-header-lean>

...
```

### Additional information:

* **Angular and @sbb-esta/angular Versions** - angular v14.0.5 with @sbb-esta/angular v14.1.1
* **Motivation for or Use Case** - We want to imitate the behavior of the header search field on [www.sbb.ch](https://www.sbb.ch/): when entering a search expression in the header
  * the page will show the list of search results, with another more central search field displayed on the page content.
  * the header search field should toggle back to the small representation (displayed as a button without a text field). For that, we tried to use the `close()` method of the `SbbHeaderSearch` directive.
* **Browsers and Operating System** - tested only on windows, with Edge Chromium, Firefox, and Chromium
* **Workarounds**:
  * **What does not work**: focusing / clicking another HTML Element of our page does unfortunately not work, as the overlay seems to be in front of all other elements.
  * **What does work**: our current workaround is, programmatically clicking on the `div.cdk-overlay-backdrop` element, so that the user does not have to do it himself.
    `document.querySelector<HTMLElement>('.cdk-overlay-backdrop')?.click();`

mhaertwig:
Closes #1579

### DIFF
--- a/src/angular/i18n/xlf/messages.xlf
+++ b/src/angular/i18n/xlf/messages.xlf
@@ -158,7 +158,7 @@
         <source>Search</source>
         <context-group purpose="location">
           <context context-type="sourcefile">../../../../../../src/angular/search/header-search.ts</context>
-          <context context-type="linenumber">81</context>
+          <context context-type="linenumber">82</context>
         </context-group>
         <note priority="1" from="description">Button label for the header search</note>
       </trans-unit>
--- a/src/angular/i18n/xlf2/messages.xlf
+++ b/src/angular/i18n/xlf2/messages.xlf
@@ -167,7 +167,7 @@
     </unit>
     <unit id="sbbSearchHeaderButtonLabel">
       <notes>
-        <note category="location">../../../../../../src/angular/search/header-search.ts:81</note>
+        <note category="location">../../../../../../src/angular/search/header-search.ts:82</note>
         <note category="description">Button label for the header search</note>
       </notes>
       <segment>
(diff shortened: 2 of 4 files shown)


kyubisation (approved):
LGTM 👍 
Nice work!

PR #1589, opened 2022-07-13
fix(angular/autocomplete): do not reset scroll if an option is added

### LINKED ISSUE #1571: Bug: Scrollbar resetting to top after loading new data in sbb-autocomplete
**Describe the bug**
When adding additional options to a sbb-autocomplete the scrollbar is reset to the beginning of all options.

**To Reproduce**
Here's a quick demonstration of the bug:
https://angular-ivy-tvz4mr.stackblitz.io/

1. Click on the input-field
2. Scroll to the bottom of the autocomplete
3. See how the scollbar resets

Demonstation only adds data once, but the bug appears everytime new data is added. To reproduce again refresh the page.

**Expected behavior**
The new data is added after the previously last element and the scrolling stays at the same position.

**Desktop:**
 - OS: Windows
 - Browser: Chromium
 - Version: 103.0.5060.114

mhaertwig:
Closes #1571

Fix to avoid scrolling to the top if a new element is added. This happens because our overlay has a dynamic height (different from @angular/material).

When the _resetBoundingBoxStyles() function in the FlexibleConnectedPositionStrategy is called, the element is reset to full height and scaled back afterwards. When scaling back the scroll position gets lost.
image

An alternative to this hack would be to either use a maxHeight property for our overlay or to fix this in angular/components.

### DIFF
--- a/src/angular/autocomplete/autocomplete-trigger.ts
+++ b/src/angular/autocomplete/autocomplete-trigger.ts
@@ -609,7 +609,18 @@ export class SbbAutocompleteTrigger
               this._changeDetectorRef.detectChanges();
 
               if (this.panelOpen) {
+                /**
+                 * Fix to avoid scrolling to the top if a new element is added.
+                 * This happens because our overlay hasa dynamic height (different from @angular/material).
+                 * When the _resetBoundingBoxStyles() function in the FlexibleConnectedPositionStrategy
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍 
(Apart from comment style 😄 )

PR #1591, opened 2022-07-14
fix(angular/lightbox): fix height calculation on mobile

### LINKED ISSUE #1586: Bug: Lightbox Buttons bugged on mobile devices
**Describe the bug**
When testing our application with mobile devices we discovered that lightbox action buttons are not or just partly visible/cut off. To exclude that this is app specific, we tested the examples at angular.app.sbb.ch and have seen the issues.
Besides the cutoff it seems that the button positioning on lightbox with confirmation is also not optimised - dont know if this is relevant for this bug.

**To Reproduce**
1. Grab your mobile
2. go to https://angular.app.sbb.ch/angular/components/lightbox/examples
3. test any lightbox
4. See cutoff and co

**Screenshots**

iphone
android_sony
android_sony confirm

**Smartphone (please complete the following information):**
 - Device: iPhone and Sony with Android

mhaertwig:
Use `window.innerHeight` instead of `100vh` to calculate the height of the `SbbLightBox`.  This is necessary because on mobile Chrome and Safari, `100vh` includes the address bar and is therefore taller than the actual viewport (see e.g. [this explanation](https://bugs.webkit.org/show_bug.cgi?id=141832#c5)). This led to cut off buttons in the lightbox.

Closes #1586

### DIFF
--- a/src/angular/lightbox/lightbox-container.ts
+++ b/src/angular/lightbox/lightbox-container.ts
@@ -10,6 +10,7 @@ import {
   HostListener,
   Inject,
   NgZone,
+  OnInit,
   Optional,
   ViewEncapsulation,
 } from '@angular/core';
@@ -40,9 +41,10 @@ import { sbbLightboxAnimations } from './lightbox-animations';
     '[attr.aria-label]': '_config.ariaLabel',
     '[attr.aria-describedby]': '_config.ariaDescribedBy || null',
     '[@lightboxContainer]': `_getAnimationState()`,
+    '[style.height.px]': '_innerHeight',
   },
 })
-export class SbbLightboxContainer extends _SbbDialogContainerBase {
+export class SbbLightboxContainer extends _SbbDialogContainerBase implements OnInit {
   /** Callback, invoked whenever an animation on the host completes. */
   @HostListener('@lightboxContainer.done', ['$event'])
   _onAnimationDone({ toState, totalTime }: AnimationEvent) {
@@ -64,6 +66,11 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
     }
   }
 
+  @HostListener('window:resize', ['$event'])
+  onResize() {
+    this._innerHeight = window.innerHeight;
+  }
+
   /** Starts the dialog exit animation. */
   _startExitAnimation(): void {
     this._state = 'exit';
@@ -72,6 +79,8 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
     // view container is using OnPush change detection.
     this._changeDetectorRef.markForCheck();
   }
+
+  _innerHeight: number;
 
   constructor(
     elementRef: ElementRef,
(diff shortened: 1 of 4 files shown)


src/angular/lightbox/lightbox-container.ts
```diff
@@ -64,6 +66,11 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
     }
   }
 
+  @HostListener('window:resize', ['$event'])
```
kyubisation, line 69: Ideally we could use the ScrollDispatcher from the CDK here, but this should work.

src/angular/lightbox/lightbox-container.ts
```diff
@@ -64,6 +66,11 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
...
+  onResize() {
```
kyubisation, line 70: Nitpick
```suggestion
  _handleResize() {
```

src/angular/lightbox/lightbox-container.ts
```diff
@@ -95,4 +104,8 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
       focusMonitor
     );
   }
+
+  ngOnInit(): void {
+    this._innerHeight = window.innerHeight;
```
kyubisation, line 109: This would break server side rendering, where the window (or document) object might not be available.
See e.g. https://github.com/sbb-design-systems/sbb-angular/blob/main/src/angular/autocomplete/autocomplete-trigger.ts#L300-L304

src/angular/lightbox/lightbox-container.ts
```diff
@@ -95,4 +104,8 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
       focusMonitor
     );
   }
+
+  ngOnInit(): void {
```
kyubisation, line 108: Could this not be calculated in the constructor?

kyubisation (changes requested):
Thank you for your work 👍

src/angular/lightbox/lightbox-container.ts
mhaertwig: I've used the `ViewportRuler`.

src/angular/lightbox/lightbox-container.ts
```diff
@@ -82,6 +89,7 @@ export class SbbLightboxContainer extends _SbbDialogContainerBase {
     ngZone: NgZone,
     overlayRef: OverlayRef,
     private _changeDetectorRef: ChangeDetectorRef,
+    private _viewportRuler: ViewportRuler,
```
mhaertwig, line 92: Seems like nobody is using the `SbbLightboxContainer` directly. Should we mark it as breaking change anyways?

kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1602, opened 2022-07-18
refactor(angular/schematics): reuse standalone component utilities

mhaertwig:
The standalone utilities were copied into `@schematics/angular` so that they can be reused.


kyubisation (approved):
LGTM 👍

PR #1616, opened 2022-07-25
fix(angular/autocomplete): prevent outside clicks from going to other overlays

mhaertwig:
Adds a subscription to the outside click stream of the autocomplete overlay so that the events don't propagate to other overlays which may close the autocomplete prematurely.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1618, opened 2022-07-25
fix(multiple): do not override styles of SBB.ch

### LINKED ISSUE #1612: Bug: Specificity of sbb-angular library overwrites specificity of other libraries
If one uses another component library besides the sbb-angular library the latter overwrites the css declarations of some elements due to a higher specificity by using `:is()` in combination with global selectors, e.g. `ul`. 

At least the following tags are affected: 
`h1-h6, a,p,ul,ol,li`

**Expected behavior**
specificity should not be overwritten

The problem was already discussed with @kyubisation and the first solution which came to mind was to combine the selectors with a :not() pseudoclass

Example: :**`is(p, ul, ol, li, a, h1, h2, h3, h4, h5, h6):not([class*='mod_']) {}`**

I haven't tested this behaviour. 

**Additional context**
angularLibv13

mhaertwig:
Closes #1612

### DIFF
--- a/src/angular/styles/_typography.scss
+++ b/src/angular/styles/_typography.scss
@@ -629,7 +629,7 @@ textarea {
 // Headings
 // ----------------------------------------------------------------------------------------------------
 
-:is(h1, .sbb-headline1) {
+:is(h1:where(:not([class*='mod_'])), .sbb-headline1) {
   font-family: var(--sbb-font-thin);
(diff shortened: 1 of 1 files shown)


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1623, opened 2022-07-26
fix(angular/processflow): remove obsolete aria-expanded attribute

### LINKED ISSUE #1613: Bug: sbb-select accessibility issue with Safari & macOs
**Describe the bug**
There is an accessibility issue with the Safari browser on macOs. 

The sbb-select does not read out with the screen reader VoiceOver the entries selected or ticked by arrow key. 

**To Reproduce**
Use macOs 15.4 or 15.5 with Safari 15.5

1, Start VoiceOver (Settings->Accessibility)
2. Go to https://angular.app.sbb.ch/angular/components/select/examples
3. Scroll down to 'Select Option Groups Multi Selection'
4. Tab to 'sbb-select International food'
5. Open sbb-select
6. Try select some option ex. Pasta
7. Error: Listen missing sound output 

**Expected behavior**
Same sound output like Chrome/Firefox on WIndows.

**Desktop**
 - OS: macOs 15.5
 - Browser safari
 - Version 15

mhaertwig:
The `aria-expanded` attribute is not valid for the `tabpanel` role (see e.g. [here](https://www.w3.org/TR/wai-aria-1.2/#tabpanel)). It can be simply removed because selecting the tab with `aria-selected` in combination with an `aria-controls` pointing to the `tabpanel` is already enough.

Closes #1613

### DIFF
--- a/src/angular/processflow/processflow.html
+++ b/src/angular/processflow/processflow.html
@@ -35,7 +35,6 @@
     (@stepTransition.done)="_animationDone.next($event)"
     [id]="_getStepContentId(i)"
     [attr.aria-labelledby]="_getStepLabelId(i)"
-    [attr.aria-expanded]="selectedIndex === i"
   >
     <ng-container [ngTemplateOutlet]="step.content"></ng-container>
   </div>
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍 
Should we maybe also open a PR at angular/components?

jeripeierSBB (approved):
LGTM

PR #1624, opened 2022-07-26
fix(multipe): don't use div containers inside labels

### LINKED ISSUE #1619: Bug: radiobutton panel not correct in html usage
**Describe the bug**
The content of a radiobutton panel content is wrapped in <label><div>...</div></label
According to HTML Standard div is not allowed in labels. 
image

**Expected behavior**
If possible you should consider using spans with block style. Check with W3C Validator or similar.

mhaertwig:
According to the specification, div containers are not allowed inside labels.

Closes #1619

### DIFF
--- a/src/angular/checkbox-panel/checkbox-panel.html
+++ b/src/angular/checkbox-panel/checkbox-panel.html
@@ -18,7 +18,7 @@
     (click)="_onInputClick($event)"
   />
 
-  <div class="sbb-selection-container">
+  <span class="sbb-selection-container">
     <span class="sbb-selection-container-checked">
       <svg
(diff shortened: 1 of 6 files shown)


kyubisation (approved):
LGTM 👍 
Should we maybe add a note to the accessibility considerations of the panels to avoid block elements?

mhaertwig:
> LGTM 👍 Should we maybe add a note to the accessibility considerations of the panels to avoid block elements?

Good idea, yes. I've updated the docs and added some tests to ensure `div`'s don't come back.

jeripeierSBB (approved):
LGTM

PR #1647, opened 2022-08-10
build: limit max memory of the bazel java process

mhaertwig:
The problem 'Worker process did not return a WorkResponse' seems to come over a skyrocketing memory usage of the java bazel process. With this limit goes the build faster and make no errors.

Refs: https://github.com/angular/angular/issues/46965

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,12 @@
+##################
+# Startup config #
+##################
+
+# The memory usage of the java process are skyrocketing with journey-maps build (bazel bug?).
+# Limit it to 2g make the build very fast and without error.
+# It's related with https://github.com/angular/angular/issues/46965
+startup --host_jvm_args=-Xmx2g
+
 ###############################
 # Filesystem interactions     #
 ###############################
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,7 +1,8 @@
-#########
-# Tests #
-#########
-
+##############
+# RAM Issues #
+##############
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -1,3 +1,12 @@
+##################
+# Startup config #
+##################
+
+# The memory usage of the java process are skyrocketing with journey-maps build (bazel bug?).
+# Limit it to 2g make the build very fast and without error.
+# It's related with https://github.com/angular/angular/issues/46965
+startup --host_jvm_args=-Xmx2g
```
julienje, line 8: Can you test the build without this?

.bazelrc
```diff
@@ -1,11 +1,6 @@
-##################
-# Startup config #
-##################
-
-# The memory usage of the java process are skyrocketing with journey-maps build (bazel bug?).
-# Limit it to 2g make the build very fast and without error.
-# It's related with https://github.com/angular/angular/issues/46965
-startup --host_jvm_args=-Xmx2g
+#########
+# Tests #
```
julienje, line 2: # RAM Issues #
...

# To prevent 'Worker process did not return a WorkResponse' maybe caused by a high memory consumption the recommandation on https://bazel.build/docs/memory-saving-mode is tested
# It's related with https://github.com/angular/angular/issues/46965

mhaertwig:
Closing, since this does not solve the build issue.
Instead a workaround is used. See #1626

PR #1652, opened 2022-08-17
ci: allow renovate to update @types/geojson and rxjs

mhaertwig:
Ignoring them was done due to a build bug (see #1609). However, it turned out they're not responsible for this bug.


jeripeierSBB (approved):
LGTM

PR #1658, opened 2022-08-24
fix(angular/lightbox): correct dependency tree for CDK dialog

### LINKED ISSUE #1656: Bug: NullInjectionException when using LightBox
**Describe the bug**
When we use the LightBox compontent in a custom component, the imports specified in @NgModule did not need 
`SbbDialogModule` under the imports section.
Since version 14 this is needed.

**Expected behavior**
The migration script (13 to 14) should add the missing import, if possilbe or at least an information in the "How to update" section would be nice.
Or even better - no new import needed at all, and the LightBox compontent imports it's needs itself.

**Error message**
The error message i got started like this:
```
Error: Uncaught (in promise): NullInjectorError: R3InjectorError(AppModule)[AddressValidationService -> SbbLightbox -> Dialog -> Dialog -> Dialog]: 
  NullInjectorError: No provider for Dialog!
```

jeripeierSBB:
Closes #1656

### DIFF
--- a/src/angular/lightbox/lightbox.module.ts
+++ b/src/angular/lightbox/lightbox.module.ts
@@ -4,6 +4,7 @@ import { PortalModule } from '@angular/cdk/portal';
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { SbbCommonModule } from '@sbb-esta/angular/core';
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍

PR #1662, opened 2022-08-29
build: update to TypeScript 4.8 and angular 14.2.0

mhaertwig:


package.json
```diff
@@ -166,13 +166,13 @@
     "tslint-config-prettier": "1.18.0",
     "tslint-consistent-codestyle": "1.16.0",
     "tslint-plugin-prettier": "2.3.0",
-    "typescript": "4.7.2",
+    "typescript": "~4.8.2",
```
kyubisation, line 169: Should this not be pinned?

package.json
```diff
@@ -166,13 +166,13 @@
...
     "yargs": "17.5.1"
   },
   "resolutions": {
-    "@angular/dev-infra-private/typescript": "4.7.2",
+    "@angular/dev-infra-private/typescript": "~4.8.2",
```
kyubisation, line 173: Same as above?

kyubisation (approved):
Basically LGTM 👍 
See comments about pinning

package.json
mhaertwig: That's how angular/components do it [now](https://github.com/angular/components/commit/e63390c712b8e66b81b6f026dcfd313c0fb3a682#diff-8f1a97df6c444527b85938f05a3cababdb66d0f97adb82370ab22c82910f5207L47) and since we're ignoring `typescript` in renovate I think unpinning it makes sense.

kyubisation (approved):

jeripeierSBB (approved):
LGTM (apart from lock file conflict)

PR #1665, opened 2022-08-30
fix(multiple): replace all usages of color grey with granite

mhaertwig:


kyubisation (approved):
LGTM 👍

src/angular/styles/_typography.scss
```diff
@@ -95,17 +95,13 @@
   --sbb-color-lemon: #{sbb.$sbbColorLemon};
   --sbb-color-green: #{sbb.$sbbColorGreen};
 
-  // Non-Standard Colors
-  --sbb-color-grey: #{sbb.$sbbColorGrey};
-  --sbb-color-grey-medium: #{sbb.$sbbColorGreyMedium};
-
```
jeripeierSBB, line 101: This is a breaking change for me which should we better not commit right now. Maybe also leave it here forever.

src/angular/styles/_variables.scss
```diff
@@ -84,10 +84,6 @@ $sbbColorPeach: #fcbb00;
 $sbbColorLemon: #ffde15;
 $sbbColorGreen: #00973b;
 
-// Non-Standard Colors
-$sbbColorGrey: #666666;
-$sbbColorGreyMedium: #979797;
-
```
jeripeierSBB, line 90: This is a breaking change for me which should we better not commit right now. Maybe also leave it here forever.

src/angular/styles/_typography.scss
mhaertwig: Good point, thanks! I agree with leaving it there forever

jeripeierSBB (approved):
LGTM

PR #1666, opened 2022-08-30
fix(multiple): resolve extended template diagnostic warnings

mhaertwig:
Turns the extended template diagnostic warnings into errors and resolves all the existing ones.


src/angular/menu/menu.html
```diff
@@ -8,8 +8,8 @@
     [@transformMenu]="_panelAnimationState"
     (@transformMenu.start)="_onAnimationStart($event)"
     (@transformMenu.done)="_onAnimationDone($event)"
-    [style.--sbb-menu-trigger-width.px]="triggerContext?.width"
-    [style.--sbb-menu-trigger-height.px]="triggerContext?.height"
+    [style.--sbb-menu-trigger-width.px]="triggerContext ? triggerContext.width : 0"
+    [style.--sbb-menu-trigger-height.px]="triggerContext ? triggerContext.height : 0"
```
jeripeierSBB, line 12: is zero a good value as fallback? should it maybe be "auto"? Just a thought

mhaertwig: These css variables are used for setting the `min-height` and `min-width` on the user menu so I think so to my understanding it does not matter.

jeripeierSBB (approved):
LGTM

PR #1675, opened 2022-09-12
fix(journey-maps/web-component): export the web-component 

swiss-chris:
Refs: ROKAS-1137

### DIFF
--- a/src/journey-maps/package.json
+++ b/src/journey-maps/package.json
@@ -31,5 +31,12 @@
   "sideEffects": false,
   "publishConfig": {
     "access": "public"
+  },
+  "exports": {
+    "./web-component": {
+      "types": "./index.d.ts",
+      "es2015": "./web-component/bundle-es2015.min.js",
+      "default": "./web-component/bundle.min.js"
+    }
   }
 }


mhaertwig (approved):
Please don't write the internal Ref in the commit message.
Otherwise LGTM 👍

kyubisation (approved):
LGTM 👍

swiss-chris:
> Please don't write the internal Ref in the commit message. Otherwise LGTM 👍

Is there any way to include the JIRA story number in the git history ?

sbboli (approved):

mhaertwig:
> > Please don't write the internal Ref in the commit message. Otherwise LGTM 👍
> 
> Is there any way to include the JIRA story number in the git history ?

No, we usually don't do it. I think the PR and the code should be self-explanatory.

swiss-chris:
> > > Please don't write the internal Ref in the commit message. Otherwise LGTM 👍
> > 
> > 
> > Is there any way to include the JIRA story number in the git history ?
> 
> No, we usually don't do it. I think the PR and the code should be self-explanatory.

Playing devil's advocate: what's the disadvantage of including the story number in the commit message (under refs) ?
The advantage: in certain cases, e.g. when fixing a bug, it might require a lot of explanations of what exactly was the bug that this commit fixes, and it would be easier to include that detail, including photos, in a jira ticket rather than the commit message.

mhaertwig:
Okay, seems that I was too strict. 😉 I discussed this internally and according to @kyubisation it's okay if the reference is in the commit body.

PR #1677, opened 2022-09-19
fix(angular/button): add correct class if icon is not a direct descendant

mhaertwig:
The `sbb-icon-button` class was not added if the icon is not a direct descendant, e.g. because it's wrapped in a link. This PR fixes this bug.

### DIFF
--- a/src/angular/button/button.spec.ts
+++ b/src/angular/button/button.spec.ts
@@ -489,7 +489,7 @@ describe('SbbButton', () => {
         )!;
 
         expect(nonIconButtonElements.length).toEqual(10);
-        expect(iconButtonElements.length).toEqual(8);
+        expect(iconButtonElements.length).toEqual(9);
 
         expect(
           nonIconButtonElements.every(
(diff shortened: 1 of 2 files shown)


src/angular/button/button.spec.ts
```diff
@@ -684,6 +684,9 @@ class ButtonIconTest extends ButtonTestBase {}
       <a sbb-alt-button type="button"><sbb-icon svgIcon="example"></sbb-icon></a>
       <a sbb-secondary-button type="button"><sbb-icon svgIcon="example"></sbb-icon></a>
       <a sbb-ghost-button type="button"><sbb-icon svgIcon="example"></sbb-icon></a>
+      <button sbb-button type="button">
```
kyubisation, line 687: Is this valid HTML?

mhaertwig: Yes, but makes no sense. I removed it.

mhaertwig: No, you're right. I didn't know this.
image

This bug appeared in our Tekton-UI which is why I created this PR. I close it and write a comment there.

kyubisation: I was vaguely aware that there was some restriction and wanted to be sure.
Thanks for checking! 👍

PR #1689, opened 2022-10-25
fix(angular/dialog): aria-modal not being set

mhaertwig:
Fixes that `aria-modal` wasn't being set on the dialog container. This likely got broken when the dialog was changed to be based on top of the CDK dialog.

The `ariaModal` option is added to the config to avoid that it's overridden by the CDK dialog (see [cdk/dialog/dialog-container.ts#L62](https://github.com/angular/components/blob/main/src/cdk/dialog/dialog-container.ts#L62))


kyubisation (approved):
LGTM 👍 
Thanks!

jeripeierSBB (approved):
LGTM

PR #1690, opened 2022-10-25
fix(angular/tooltip): animations running when timeouts haven't elapsed

mhaertwig:
After the tooltips were switched to CSS animations, a regression was introduced where the opposite animation is shown even if the tooltip didn't actually reach its target state.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1695, opened 2022-10-27
fix(angular/radio): radio not preselected with static value and an ngIf

mhaertwig:
Fixes that radio buttons weren't being preselected correctly when they have an `ngIf` and a static `value`. It appears that the issue was because we were subscribing to the `UniqueSelectionDispatcher` before the name of the radio button is assigned.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1697, opened 2022-10-27
build: account for new dev-infra packages @angular/ng-dev and `@angular/build-tooling`

mhaertwig:
The dev-infra-private packages were split to decouple the ng-dev tool from the build tooling. This allows to easily update ng-dev e.g. in LTS branches without needing to update Bazel.

It is important that ng-dev can be updated as much as possible in all branches to ensure the release IPC reliably works.


kyubisation (approved):
LGTM 👍

PR #1703, opened 2022-11-02
build: specify target versions vor schematics, update to rc-1

mhaertwig:
BREAKING CHANGE: Since `@angular/cdk/schematics` removed old migrations from their repository, we specify our own target versions.


src/angular/schematics/ng-update/target-version.ts
```diff
@@ -0,0 +1,6 @@
+/** Possible versions that can be automatically migrated by `ng update`. */
+export enum TargetVersion {
```
jeripeierSBB, line 2: Don't we have access to the original TargetVersion anymore?

kyubisation (approved):
LGTM 👍 
Should we maybe just remove the old migrations?

mhaertwig:
> LGTM 👍 Should we maybe just remove the old migrations?

Yes!! Your answer to this question in Teams was obvious but I misunderstood it. 😉 My fault. I also prefer removing them.

PR #1704, opened 2022-11-02
fix(angular/tooltip): change tooltip icon to info

### LINKED ISSUE #1684: Tooltip default icon not aligned with Styleguide
Currently the default icon of the tooltip is "questionmark", but https://digital.sbb.ch/de/webapps/components/tooltip says the default icon should be "i" (info). 

This should be aligned and the default icon should be changed to "i".

**Describe alternatives you've considered**
None.

**Additional context**
This inconsistency lead to the situation where we mixed both icons in our application and now have to clean it up.

mhaertwig:
BREAKING CHANGE: We align the default tooltip icon with the SBB style guide and replace the question mark with an info icon.

Closes #1684

### DIFF
--- a/src/angular/tooltip/tooltip-wrapper.html
+++ b/src/angular/tooltip/tooltip-wrapper.html
@@ -11,7 +11,7 @@
   #tooltip="sbbTooltip"
   (click)="tooltip.show(0)"
 >
-  <sbb-icon [svgIcon]="svgIcon"></sbb-icon>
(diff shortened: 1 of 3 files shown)


jeripeierSBB (commented):
I'm not sure if we really should change that? @kyubisation

kyubisation (changes requested):
In standard it should be `kom:circle-question-mark-small` and in lean it should be `kom:circle-information-small`
Compare with https://digital.sbb.ch/

kyubisation (approved):
LGTM 👍

PR #1713, opened 2022-11-08
refactor: remove old migrations

mhaertwig:
BREAKING CHANGE: Remove old `ng-update` migrations because sbb-angular packages cannot be updated across multiple versions.


kyubisation (approved):
LGTM 👍 
Thanks! 😃

PR #1718, opened 2022-11-15
feat(angular/select): add page down/up button functionality

mhaertwig:
Implement page up and down functionality for select:
* Page-Down focuses the last element in the select
* Page-Up focus the first element in the select
* It's done in open and closed state


jeripeierSBB (approved):
LGTM

PR #1720, opened 2022-11-16
fix(multiple): clean up list key manager on destroy

mhaertwig:
Historically we haven't needed to do any cleanup for the `ListKeyManager`, because it didn't contain any rxjs subscriptions. Over time more functionality was accumulated which now needs to be cleaned up.

These changes call the new `destroy` method in all the relevant places.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1723, opened 2022-11-17
chore(deps): update angular to v15

mhaertwig:

PR #1727, opened 2022-11-17
fix(angular/select): don't assign typeahead value after blur

mhaertwig:
Fixes that the select would assign the value triggered by the typeahead even if the select doesn't have focus anymore.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM apart from builds ;-)

mhaertwig:
> LGTM apart from builds ;-)

Yes, it needs the new CDK so https://github.com/sbb-design-systems/sbb-angular/pull/1723 has to be merged first. I am working on it :-)

PR #1739, opened 2022-11-22
ci: add lean polyfill for testing

mhaertwig:
`@angular/schematics` does not provide a `test.js` file by default anymore. Instead of configuring the lean design variant in the `test.js` file, we add a polyfill for doing this.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍 
Nice work!

PR #1740, opened 2022-11-23
chore: adjust ci workflow after release

mhaertwig:


.github/workflows/maintenance-changelog.yml
```diff
@@ -3,6 +3,7 @@ name: cherry-pick changelog from release branch into default branch
 on:
   push:
     tags: ['*.*']
+    branches: ["*.x"]
```
jeripeierSBB, line 6: why is this change needed?

mhaertwig: Because otherwise this action tries to cherry-pick major releases (on main) into release branches (e.g. 15.x). However, we only want to allow it the other direction.

jeripeierSBB (approved):
LGTM

PR #1741, opened 2022-11-23
build: update angular and cdk version range

mhaertwig:


jeripeierSBB (approved):
LGTM, is it maybe missing in release docu?

jeripeierSBB (approved):
LGTM

PR #1750, opened 2022-12-01
ci: fix renovate by setting `skipInstalls` to `false`

mhaertwig:
`psstUpgradeTasks` using `yarn` stopped working if `skipInstalls` is set to `true`.

Maybe caused by https://github.com/renovatebot/renovate/blob/main/lib/modules/manager/npm/post-update/lerna.ts#L63

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -26,6 +26,7 @@
     "yarn",
     "zone.js"
   ],
+  "skipInstalls": false,
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
Nice, and thank you!! I hope it will fix it.

PR #1752, opened 2022-12-01
fix(angular/menu): explicitly set aria-expanded to true/false

mhaertwig:
In the context of a submenu trigger it is useful for a screen reader user to know whether a submenu is open or closed. By specifying the `aria-expanded` attribute as either true or false it provides an additional "collapsed" announcement to the user informing them of the state of the menu trigger when closed.


jeripeierSBB (approved):
LGTM

PR #1753, opened 2022-12-01
fix(angular/select): changed after checked error if option label changes

mhaertwig:
Fixes a "changed after checked" error that is thrown if the label of a selected option changes dynamically.


kyubisation (approved):
LGTM 👍

PR #1754, opened 2022-12-01
refactor(angular/autocomplete): declare injected properties as nullable

mhaertwig:
In autocomplete-trigger.ts, add `| null` to the typing of optionally injected properties. This improves code health because the compiler can detect possible type errors with `null`.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1781, opened 2023-01-16
fix(angular/tabs): allow focus on disabled tabs

mhaertwig:
According to the WCAG best practices focus should be allowed on disabled tabs (see https://w3c.github.io/aria-practices/#kbd_disabled_controls). These changes update our tabs to be in line with the recommendation.


mhaertwig:
I mark this PR as draft because currently it's not clear how to style tabs that are disabled and focussed.
Moreover, in the current solution active and focussed tabs are not highlighted either.

kyubisation:
We probably need to discuss this with @mcilurzo

PR #1787, opened 2023-01-17
feat(angular/datepicker): add option to prevent entering overflowing dates

### LINKED ISSUE #1761: Bug: Overflowing date in datepicker
**Describe the bug**
In a datepicker, one can enter (keyboard) non-existing dates that results in different (valid) dates. 

**To Reproduce**
1. Go to https://angular.app.sbb.ch/angular/components/datepicker/examples
2. Enter the text "30.02.2023" into the first input field "Datepicker"
3. Bug A: Observe the model value (displayed below): "02.03.23"
4. Leave the field
5. Bug B: The text in the field has changed to "Th, 02.03.2023".

Other examples:
1. 57.15.2000 => 26.04.2001
2. 00.00.2000 => 30.11.1999

StackBlitz (link from example above)
https://stackblitz.com/run?file=src%2Fapp%2Fdatepicker-simple-reactive-example.ts

**Expected behavior**
Bug A: The model value is null as for other inputs that are not valid dates (e.g. "text")
Bug B: The text in the field does not change. It remains as entered "30.02.2023".

mhaertwig:
Closes #1761


src/angular/core/datetime/native-date-adapter.ts
```diff
@@ -210,17 +219,35 @@ export class SbbNativeDateAdapter extends SbbDateAdapter<Date> {
     return !isNaN(date.valueOf());
   }
 
+  isOverflowingDate(value: string): boolean {
```
kyubisation, line 222: Is it necessary for this method to be public? I would prefer it, if it was private or at least protected.

kyubisation (changes requested):
One nitpick, other than that it looks fine 👍

kyubisation (approved):
LGTM 👍 
Thanks!

PR #1804, opened 2023-01-31
fix(angular/dialog): scrollable content if no dialog height is defined

### LINKED ISSUE #1803: Bug: SbbDialog not diaplayed correctly
**Describe the bug**
we use the sbbDialog component with content and an action bar. 
when the browser window reaches a particular height, the action bar is no longer properly displayed (look at attached image)

**To Reproduce**
1. go to https://angular.app.sbb.ch/angular/components/dialog/examples
2. open the Dialog in the 3rd example
3. make the browser window smaller (height) until the button start to disapear

**Expected behavior**
the action bar is properly displayed no matter the height of the window
the content of the dialog is scrollable

**Screenshots**
image

image

OS: Windows
Browser: chrome (109.0.5414.120) and firefox (109.0)
Version of Angular 14.0.4,
Version of SBB Library: 14.1.3

mhaertwig:
If no dialog height is specified, the  content was not scrollable and the dialog footer sometimes was not visible. This is because flexbox uses `min-height: auto` by default. To make the dialog content scrollable again, the `min-height` is set to zero.

Closes #1803

### DIFF
--- a/src/angular/dialog/dialog.scss
+++ b/src/angular/dialog/dialog.scss
@@ -37,6 +37,10 @@
   height: 100%;
   display: flex;
   flex-direction: column;
+
+  // Flexbox uses `min-height: auto` by default.
+  // To make the content scrollable, we set it to `0`.
+  min-height: 0;
 }
 
 .sbb-dialog-content {


kyubisation (approved):
LGTM 👍 
Nice find!

PR #1810, opened 2023-02-07
ci: fix publishing of npm package

mhaertwig:
This reverses a change from #1805 that caused the NPM package to not be published.


kyubisation (approved):
LGTM 👍

PR #1813, opened 2023-02-08
feat(journey-maps): display bounding box info

besartmemeti:


src/components-examples/journey-maps/angular/journey-maps-full/journey-maps-full-example.ts
```diff
@@ -322,6 +327,18 @@ export class JourneyMapsFullExample implements OnInit, OnDestroy {
     return this.mapCenter as { lng: number; lat: number };
   }
 
+  mapBoundingBoxInfo(): { lngSW: number; latSW: number; lngNE: number; latNE: number } | undefined {
+    if (!this.mapBoundingBox) {
+      return;
+    }
+    return {
```
sebisteiner, line 334: Wouldn't it be easier to directly return a formatted string here?

sebisteiner (changes requested):

sebisteiner (approved):
LGTM 👍

src/showcase/app/journey-maps/journey-maps.module.ts
```diff
@@ -1,5 +1,5 @@
 import { PortalModule } from '@angular/cdk/portal';
-import { CommonModule } from '@angular/common';
+import { CommonModule, DecimalPipe } from '@angular/common';
```
kyubisation, line 2: DecimalPipe is already provided by CommonModule: https://github.com/angular/angular/blob/15.1.4/packages/common/src/pipes/index.ts#L55
This change is unnecessary.

kyubisation (changes requested):
See comment above

src/showcase/app/journey-maps/journey-maps.module.ts
besartmemeti: It still has to be listed as a provider. The following error occurs when it is not given

image

sebisteiner:
> See comment above

I solved it without using the `DecimalPipe`. Can you please re-review?

kyubisation (approved):
LGTM 👍 
I just realized, I initially misread the code. You were right to register it as a provider to inject it via DI.
I apologize for that misunderstanding.
`toFixed` however seems to be the more efficient solution here.

PR #1814, opened 2023-02-08
feat(journey-maps): switch between 2D and 3D layers

swiss-chris:
depdnging on the levelSwitch uiOptions

Refs: ROKAS-1078


src/journey-maps/angular/services/map/two-d-three-d-service.ts
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * If the level-switch feature is enabled by the client, then show only 3d layers (-lvl)
+   * If the level-switch feature is disabled by the client, then show only 2d layers (-2d)
+   */
+  public show2Dor3D(map: MaplibreMap, isLevelSwitchEnabled: boolean) {
+    this._setVisibility(map, '-2d', isLevelSwitchEnabled ? 'none' : 'visible');
+    this._setVisibility(map, '-lvl', isLevelSwitchEnabled ? 'visible' : 'none');
+  }
+
+  private _setVisibility(map: MaplibreMap, searchString1: string, newVisibility1: string) {
```
sebisteiner, line 17:
```suggestion
  private _setVisibility(map: MaplibreMap, layerIdSuffix: string, visibility: 'visible' | 'none') {
```

sebisteiner (dismissed):

besartmemeti:
LGTM

src/journey-maps/angular/services/map/two-d-three-d-service.ts
swiss-chris: Done ✅

sebisteiner: Don't you like `visibility: 'visible' | 'none'` 🥲

sebisteiner (approved):
LGTM 👍

src/journey-maps/angular/services/map/two-d-three-d-service.ts
swiss-chris: Sorry, das hatte ich nicht gesehen. Hab's gepushed.

src/journey-maps/angular/services/map/two-d-three-d-service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+import { Map as MaplibreMap } from 'maplibre-gl';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class TwoDThreeDService {
```
kyubisation, line 7: Out of curiosity; Does this functionality really need a separate class for it? The overhead for bundling, runtime and just developer mindspace seems barely worth it?

swiss-chris: As it currently stands, this service is only called from one single method, so yes, we could inline this code to the same class as the method currently calling it. Let me know what you prefer and I can refactor it in a few minutes.

swiss-chris: In general, what kind of criteria do/would you use to decide whether or not to create a separate service ?

kyubisation: As usual, it's not exactly obvious what the right decision is.
I think in this specific case, I would prefer to inline this functionality.
From my perspective, there is no merit in separating this code into its own class.
But as mentioned, YMMV.

swiss-chris: I'll push the change in a minute after bazel lets me test it :)

swiss-chris: > I'll push the change in a minute after bazel lets me test it :)

https://xkcd.com/303/

swiss-chris: > I'll push the change in a minute after bazel lets me test it :)

Done ✅

kyubisation (approved):
LGTM 👍

src/journey-maps/angular/services/map/two-d-three-d-service.ts
kyubisation: > In general, what kind of criteria do/would you use to decide whether or not to create a separate service ?

I'm afraid I don't have a clear answer.
There is the difference between library and product development. In product development you are far more free in structuring your app, but in library development you have to pay attention to a lot more of attention (e.g. performance on component level and every separate class is a slight overhead).
So, generally, do what feels right. But I might ask questions if I have doubts. 😃

PR #1847, opened 2023-03-21
fix(angular/core): sbb-option sets aria-selected="false"

mhaertwig:
For sbb-option, set `aria-selected="false"` on deselected options. Confirms with [WAI ARIA Listbox authoring practices guide](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/), which says to always include aria-selected attribute on listbox options that can be selected. Fix issue where VoiceOver reads every option as "selected".

### DIFF
--- a/src/angular/core/option/option.ts
+++ b/src/angular/core/option/option.ts
@@ -56,15 +56,6 @@ export class SbbOptionSelectionChange<T = any> {
     '[class.sbb-option-multiple]': 'multiple',
     '[class.sbb-focused]': 'active',
     '[id]': 'id',
-    // Set aria-selected to false for non-selected items and true for selected items. Conform to
-    // [WAI ARIA Listbox authoring practices guide](
-    //  https://www.w3.org/WAI/ARIA/apg/patterns/listbox/), "If any options are selected, each
-    // selected option has either aria-selected or aria-checked set to true. All options that are
-    // selectable but not selected have either aria-selected or aria-checked set to false." Align
-    // aria-selected implementation of Chips and List components.
-    //
-    // Set `aria-selected="false"` on not-selected listbox options to fix VoiceOver announcing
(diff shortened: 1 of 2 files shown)


src/angular/core/option/option.ts
```diff
@@ -56,7 +56,16 @@ export class SbbOptionSelectionChange<T = any> {
...
+    // Set aria-selected to false for non-selected items and true for selected items. Conform to
+    // [WAI ARIA Listbox authoring practices guide](
+    //  https://www.w3.org/WAI/ARIA/apg/patterns/listbox/), "If any options are selected, each
+    // selected option has either aria-selected or aria-checked set to true. All options that are
+    // selectable but not selected have either aria-selected or aria-checked set to false." Align
+    // aria-selected implementation of Chips and List components.
+    //
+    // Set `aria-selected="false"` on not-selected listbox options to fix VoiceOver announcing
+    // every option as "selected" (#21491).
+    '[attr.aria-selected]': 'selected',
```
jeripeierSBB, line 68: why is set to `selected`? It somehow does not align with the comment for me :D

mhaertwig: Before this fix the `aria-selected` attribute was removed on unselected options in a single select. Now it's `false` by default.

But you're right: The comment does not really make sense because it explains why we use the default behaviour and removed the workaround. It's probably more clear without this comment. :-D I copied it without questioning it too much

jeripeierSBB: it's not false, it's "selected" ?

mhaertwig: It is `selected` which is `false` by default.

jeripeierSBB: ah it's a variable not a string :D

jeripeierSBB (approved):
LGTM

PR #1848, opened 2023-03-21
fix(oauth): parse token correctly to be able to parse unicode characters

bweben:
Hello

Currently, the parsing of access tokens does not quite work. Unicode characters can generate problems.
Even so far to produce an error when calling `atob` which leads to the token not being able to be parsed at all.

This PR will address that. The logic to do that is mainly taken from https://stackoverflow.com/a/38552302 

Thanks


kyubisation (approved):
LGTM 👍 
Thanks for the contribution!

PR #1850, opened 2023-03-22
fix(angular/checkbox): fix ARIA semantics and use native DOM properties

mhaertwig:
For the checkbox component, fix semantics so that a11y produces a state of "mixed" when both the `checked` and `indeterminate` inputs to sbb-checkbox are both true.

Also, remove the use of aria-checked in favor of the native DOM properties for checked and indeterminate. Conform to [ARIA in HTML W3C input checkbox specification](https://www.w3.org/TR/html-aria/#el-input-checkbox).

Fix a11y bug where checkbox renders the indeterminate visual state but accessibility tree has a state of "checked". Aligns with native checkbox, which produces the mixed state when indeterminate property is true, regardless of the value of the checked property.

Some accessibility checkers produce an error when native checkbox has aria-checked on it. Remove aria-checked to fix errors.

Does not make visual changes.


jeripeierSBB (approved):
LGTM

PR #1866, opened 2023-04-05
fix(angular/datepicker): remove div as a child of button

mhaertwig:
Fixes that we had divs as children of buttons which some a11y tools complain about.


jeripeierSBB (approved):
LGTM

PR #1873, opened 2023-04-14
fix(journey-maps): fix multiline attribution style

sebisteiner:
Refs: ROKAS-1404

### DIFF
--- a/src/journey-maps/angular/components/attribution/attribution.html
+++ b/src/journey-maps/angular/components/attribution/attribution.html
@@ -6,9 +6,9 @@
   aria-hidden="true"
 >
   <div class="attribution-text">
-    <span *ngFor="let attribution of attributions">
-      <span [innerHTML]="attribution" class="attribution-item"></span>
-    </span>
+    <ng-container *ngFor="let attribution of attributions">
+      <div [innerHTML]="attribution" class="attribution-item"></div>
+    </ng-container>
   </div>
   <div class="attribution-button">
     <div class="map-control-container">
--- a/src/journey-maps/angular/components/attribution/attribution.ts
+++ b/src/journey-maps/angular/components/attribution/attribution.ts
@@ -4,8 +4,10 @@ import {
   Input,
   OnChanges,
   OnDestroy,
+  SecurityContext,
   SimpleChanges,
 } from '@angular/core';
+import { DomSanitizer } from '@angular/platform-browser';
 import { Map as MaplibreMap, MapDataEvent } from 'maplibre-gl';
 
 @Component({
@@ -28,7 +30,7 @@ export class SbbAttribution implements OnChanges, OnDestroy {
   private readonly _compactBreakpoint = 640;
   private readonly _linkRegex = /<a .+?<\/a>/gi;
 
-  constructor(private _cd: ChangeDetectorRef) {}
+  constructor(private _cd: ChangeDetectorRef, private _sanitizer: DomSanitizer) {}
 
   ngOnChanges(changes: SimpleChanges): void {
     const prev = changes.map?.previousValue;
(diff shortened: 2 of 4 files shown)


src/journey-maps/angular/components/attribution/attribution.ts
```diff
@@ -75,15 +77,17 @@ export class SbbAttribution implements OnChanges, OnDestroy {
     const attributions = new Set<string>();
```
kyubisation, line 77: Why not the following?
```
this.attributions = Object.values(sourceCaches)
      .filter((sc) => sc.used || sc.usedForTerrain)
      .map((sc) => sc.getSource().attribution)
      .map((attr) => this._sanitizer.sanitize(SecurityContext.HTML, attr ?? null))
      .filter((attr) => !!attr)
      .flatMap((attr) => this._splitAttribution(attr!))
      .filter((v, i, a) => a.indexOf(v) === i)
      .sort(this._sortAttributionItems);
```

sebisteiner: I implemented your suggestion

swiss-chris:
Hallo @kyubisation 
Könnt ihr euch diesen PR bei Gelegenheit nochmals anschauen ?

src/journey-maps/angular/components/attribution/attribution.ts
```diff
@@ -72,18 +74,17 @@ export class SbbAttribution implements OnChanges, OnDestroy {
...
-    for (const sourceCache of Object.values(sourceCaches)) {
-      if (sourceCache.used || sourceCache.usedForTerrain) {
-        const attribution = sourceCache.getSource().attribution;
-        this._splitAttribution(attribution).forEach((attr) => attributions.add(attr));
-      }
-    }
+    this.attributions = Object.values(sourceCaches)
+      .filter((sc) => sc.used || sc.usedForTerrain)
+      .map((sc) => sc.getSource().attribution)
+      .map((attr) => this._sanitizer.sanitize(SecurityContext.HTML, attr ?? null))
```
jeripeierSBB, line 82: should this two maps maybe be combined to avoid an extra iteration? or should the filter be before the second map so "?? null" could be removed?

src/journey-maps/angular/components/attribution/attribution.html
```diff
@@ -6,9 +6,9 @@
   aria-hidden="true"
 >
   <div class="attribution-text">
-    <span *ngFor="let attribution of attributions">
-      <span [innerHTML]="attribution" class="attribution-item"></span>
-    </span>
+    <ng-container *ngFor="let attribution of attributions">
```
jeripeierSBB, line 9: is the ng-container really needed or could "*ngFor="let attribution of attributions"" be applied to the div? I am not sure, because of the `innerHTML` if it really works :)

jeripeierSBB (approved):
LGTM

src/journey-maps/angular/components/attribution/attribution.ts
swiss-chris: Ich persönlich bevorzuge die 2 separaten Zeile weil ich es besser lesbar finde. Bezüglich Performance meint ChatGPT 4:
"There is no significant performance advantage in combining the two adjacent map() calls into one line. JavaScript engines are highly optimized, and the performance difference between these two approaches would be negligible."

Ich kann nicht abschätzen, ob das stimmt, aber ich hätte es mal so gelassen.

src/journey-maps/angular/components/attribution/attribution.html
swiss-chris: I tried your suggestion and it works:

```
    <div *ngFor="let attribution of attributions" [innerHTML]="attribution" class="attribution-item"></div>
```

But in the end I decided not to push the change as I find it more readable the way it was before.

src/journey-maps/angular/components/attribution/attribution.ts
jeripeierSBB: ist ok für mich, obwohl ich chat gpt nicht vollständig vertrauen würde :D. https://stackoverflow.com/questions/55680766/does-javascript-optimise-multiple-pure-filters-maps-etc

PR #1879, opened 2023-04-18
build: fix build dependencies

mhaertwig:
Fix build by adding `@babel/plugin-proposal-async-generator-functions` as a dev dependency. It's needed by `@angular/build-tooling`.


kyubisation (approved):
LGTM 👍 
Does this not have a yarn.lock change?

PR #1881, opened 2023-04-19
build: lint BUILD.bazel files

mhaertwig:
This is necessary to avoid linting errors when updating bazel to v6.1.0.


kyubisation (approved):
LGTM 👍

PR #1898, opened 2023-05-04
chore: prepare the 16 release

mhaertwig:
BREAKING CHANGE: This prepares the 16 release


jeripeierSBB (approved):
LGTM, but I did not think about whether there are more cases which should be covered. But I think you worked according to the guide, so it should be ok :D

mhaertwig:
I followed the guide and made some preparations in advance. Today, I'll test everything and, if everything works well, make a release.

PR #1904, opened 2023-05-22
fix(multiple): fix VoiceOver confused by Select/Autocomplete's ARIA semantics

mhaertwig:
For Select and Autocomplete components, fix issues where VoiceOver was confused by the ARIA semantics of the combobox. Fix multiple behaviors:

 - Fix VoiceOver focus ring stuck on the combobox while navigating options.
 - Fix VoiceOver would sometimes reading option as a TextNode and not communicating the selected state and position in set.
 - Fix VoiceOver "flickering" behavior where VoiceOver would display one announcement then quickly change to another announcement.

Fix the same issues for both Select and Autocomplete component.

Implement fix by correcting the combobox element and also individual options.

First, move the aria-owns reference to the overlay from the child of the combobox to the parent modal of the comobobx. Having an aria-owns reference inside the combobox element seemed to confuse VoiceOver.

Second, apply `aria-hidden="true"` to the ripple element and pseudo checkboxes on sbb-option. These DOM nodes are only used for visual purposes, so it is most appropriate to remove them from the accessibility tree. This seemed to make VoiceOver's behavior more consistent.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   "schematics": "./tools/schematics/collection.json",
   "dependencies": {
     "@angular/animations": "16.0.2",
-    "@angular/cdk": "16.0.1",
+    "@angular/cdk": "16.1.0-next.0",
     "@angular/common": "16.0.2",
     "@angular/core": "16.0.2",
     "@angular/elements": "16.0.2",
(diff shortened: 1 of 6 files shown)


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

mhaertwig:
Fresh version: #2528 2528

PR #1916, opened 2023-06-12
fix(stackblitz): fix dependencies and registry

mhaertwig:
This updates the dependency `zone.js` to 0.13 and fixes the registry url for the stackblitz assets.


kyubisation (approved):
LGTM 👍 
Thanks! 😃

PR #1924, opened 2023-06-15
build: update to typescript 5.1

mhaertwig:
Update typescript and remove obsolete babel dependencies (the build worked after deleting node_modules and clearing the bazel cache).


kyubisation (approved):
LGTM 👍

PR #1927, opened 2023-06-22
feat(journey-maps): show stack-blitz example link

mstankala:


src/components-examples/journey-maps/angular/journey-maps-basic/journey-maps-basic-example.ts
```diff
@@ -1,4 +1,6 @@
 import { Component } from '@angular/core';
+import { SbbNotificationModule } from '@sbb-esta/angular/notification';
+import { SbbJourneyMapsModule } from '@sbb-esta/journey-maps/angular/journey-maps.module';
```
mhaertwig, line 3:
```suggestion
import { SbbJourneyMapsModule } from '@sbb-esta/journey-maps';
```
@mstankala, after applying this change, you need to run `yarn baz` and commit all changes. Stackblitz will work but the map is not showing due to the missing token. Do you already have an idea how to solve this?

mstankala: @mhaertwig No, it still doesn't work.

mstankala: OK, with your suggested change, it work now. Thx 
@mhaertwig What do you suggest, how can we pass the test api-key into the Stackblitz example?

kyubisation: Hello @mstankala 
Look for `${sbbLean}` in the code for an example on how to pass data into StackBlitz

src/showcase/app/shared/component-viewer/stack-blitz/stack-blitz-writer.ts
```diff
@@ -82,6 +82,16 @@ export class StackBlitzWriter {
           isSbbLean ? ' class="sbb-lean"' : '',
         );
 
+        if (
+          files['src/app/journey-maps-basic-example.html'] ||
```
kyubisation, line 86: I would recommend to reduce this to something like `Object.keys(files).some(f => f.includes('journey-maps')`

src/showcase/app/shared/component-viewer/stack-blitz/stack-blitz-writer.ts
```diff
@@ -82,6 +82,16 @@ export class StackBlitzWriter {
...
+          files['src/app/journey-maps-full-example.html']
+        ) {
+          files['src/index.html'] = files['src/index.html'].replace(
+            '</head>',
+            "<script>window.JM_API_KEY='" + window['JM_API_KEY'] + "'</script></head>",
```
kyubisation, line 91: Prefer string templates

src/components-examples/journey-maps/angular/journey-maps-basic/journey-maps-basic-example.ts
mstankala: How can I pass all file with exmaple-data from 'src/components-examples/journey-maps/angular/journey-maps-full/mock-response/*' into StackBlitz. 
What is the best way?

kyubisation: I'm not fond of it, as this can very easily break when changes are implemented.
But I guess I can live with it.

mstankala:
> Is this ready to merge?

Yes

PR #1947, opened 2023-07-24
fix(angular/autocomplete): blocking events to other overlays when the…

mhaertwig:
…re are no results

The autocomplete trigger attaches an overlay even if there are no options in the list. It also subscribes to keydown events and clicks while the panel is open which block events from reaching other overlays. This means that if the user focuses an autocomplete input with no options, the events from it won't reach other overlays.

These changes resolve the issue by subscribing and unsubscribing from the event streams depending on the visibility state of the panel.


jeripeierSBB (approved):
LGTM

PR #1950, opened 2023-07-25
fix(angular/radio): clear selected radio button from group

mhaertwig:
In  a previous commit the radio group was changed so that deselected buttons receive `tabindex="-1"` when there's a selected button. The problem is that we weren't clearing the reference to the selected button so it gets removed, the deselected buttons become unfocusable using the keyboard.

These changes clear the selected radio button on destroy.


kyubisation (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #1952, opened 2023-07-26
feat(angular/autocomplete): add input to require selection from the panel

mhaertwig:
Adds the `requireSelection` input to the autocomplete, which when enabled will clear the input value if the user doesn't select an option from the list.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1953, opened 2023-07-26
feat(angular/header-lean): add directive to display icon buttons in header

### LINKED ISSUE #902: Feature request:
According to https://digital.sbb.ch/de/webapps/modules/header it is possible to place icons (search and bell) left of the usermenu in the header of a webapp. However, from what I can see this is not yet possible (without CSS hacks) in the current implementation of https://angular.app.sbb.ch/business/components/header 

Add a possibility to easily add those kind of icons.

mhaertwig:
Adds the `<sbb-header-icon-actions />` directive for displaying icon buttons on the right side of the header, left to the `<sbb-usermenu>` and the `[brand]` element.

Closes #902

### DIFF
--- a/src/angular/header-lean/header.scss
+++ b/src/angular/header-lean/header.scss
@@ -32,6 +32,14 @@
   }
 }
 
+.sbb-header-icon-actions {
+  white-space: nowrap;
+
+  .sbb-frameless-button:not(:last-child) {
+    margin-right: calc(#{sbb.pxToRem(15)} * var(--sbb-scaling-factor));
+  }
+}
+
 /** Header environment styles. */
 .sbb-header-environment {
   z-index: 1;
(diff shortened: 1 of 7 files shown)


src/angular/header-lean/header.scss
```diff
@@ -32,6 +32,14 @@
   }
 }
 
+.sbb-header-icon-actions {
+  white-space: nowrap;
+
+  .sbb-frameless-button:not(:last-child) {
```
kyubisation, line 38: Does this need to be this specific to `.sbb-frameless-button`?
I feel like we could maybe use `flex` with `gap` here.
What do you think?

mhaertwig: Totally makes sense. Thanks!

mhaertwig:
image

kyubisation (approved):
LGTM 👍 
Thanks! 😃

jeripeierSBB (approved):
LGTM, nice :)

PR #1972, opened 2023-08-17
fix(angular/table): resolve local compilation issues

mhaertwig:
The SBB Angular table had a few places where it was importing a component's template as a string. This is incompatible with the upcoming local compilation mode in the compiler. These changes inline the templates instead.


kyubisation (approved):
LGTM 👍

PR #1980, opened 2023-08-30
feat(angular/sidebar): implement dark mode

mhaertwig:


matthiaswattingersbb (approved):

src/angular/styles/_typography.scss
```diff
@@ -112,6 +112,9 @@
   --sbb-color-text: var(--sbb-color-black);
   --sbb-color-weak-accent: var(--sbb-color-granite);
 
+  // Dark mode colors
+  --sbb-dark-mode-color-red: #{sbb.$sbbDarkModeColorRed};
```
jeripeierSBB, line 116: In my opinion it sounds better to have `dark-mode` as suffix. What do you think?

mhaertwig: I actually had this at the beginning but for me it sounded like it was a dark red.  That's why I decided to prefix the colors with `dark-mode`. But am not committed to that. Maybe ask the others in the train what they think 😉

jeripeierSBB: @kyubisation

kyubisation: Good question.
Comparing with lean, we did not declare changed variables on their own and just override existing ones.
As another option, we could use a different separator for this, e.g. `--sbb-color-red_dark-mode`.
What do you guys think?

mhaertwig: We'll need the dark mode red in different places which is why I decided to declare a variable.
Let's go with `--sbb-color-red_dark-mode`.

kyubisation (approved):
LGTM 👍 
Nice work 😃

jeripeierSBB (approved):
LGTM

PR #1983, opened 2023-08-31
feat(multiple): implement dark mode

mhaertwig:
BREAKING CHANGE: This adds the dark mode for the lean design variant. The mode chosen is based on the system settings. To enforce a specific mode, you can add the `sbb-light` or `sbb-dark` class to the `<html>` element.


kyubisation (approved):
LGTM 👍 
Amazing work 😃

PR #1995, opened 2023-09-07
feat(angular/chips): implement dark mode

antoniafurrer:


src/angular/styles/_typography.scss
```diff
@@ -174,6 +174,11 @@
   );
   --sbb-chips-chip-padding-horizontal: calc(#{sbb.pxToRem(8)} * var(--sbb-scaling-factor));
   --sbb-chips-input-width: calc(#{sbb.pxToRem(150)} * var(--sbb-scaling-factor));
+  --sbb-chips-background-color-active: var(--sbb-color-cloud);
+  --sbb-chips-background-color-focused: var(--sbb-color-aluminum);
+  --sbb-chips-background-color-hover: var(--sbb-color-aluminum);
+  --sbb-chips-background-color-disabled: var(--sbb-color-cloud);
+  --sbb-chips-color: var(--sbb-color-weak-accent);
```
jeripeierSBB, line 181: I'm wondering wether we should not change weak accent globally as this was the original goal of this context specific variable that it is defined globally. So hopefully it would also work in dark context. @mcilurzo do you know that?


kyubisation, line 1: What is this file? I'm confused...

antoniafurrer: Same...I am going to try to somehow delete it, I guess

PR #1997, opened 2023-09-07
feat(angular/form-field): implement dark mode

antoniafurrer:


src/angular/styles/_typography.scss
```diff
@@ -1068,12 +1083,12 @@ textarea {
   }
 
   &:is(:focus, .sbb-focused) {
-    border-color: var(--sbb-color-weak-accent);
+    border-color: var(--sbb-form-border-color-active);
```
jeripeierSBB, line 1086: not sure if we should remove --sbb-color-weak-accent here? related to other comments of me :D

antoniafurrer: It is not the same color as the week accent we chose for the weak accent in darkmode so its fine that way :D

mhaertwig: @jeripeierSBB is correct:
`--sbb-color-weak-accent` has `--sbb-color-iron` in Lean mode and `--sbb-color-granite` in standard mode.
You also changed the focus color of the form field in standard mode what is not intended.

I suggest to set `--sbb-form-border-color-active` to `var(--sbb-color-weak-accent)` in line 234 and to override `--sbb-form-border-color-active` for the dark mode.

src/angular/styles/_typography.scss
```diff
@@ -228,11 +228,14 @@
   --sbb-fieldset-legend-nested-font-size: #{sbb.pxToRem(20)};
 
   // Forms
-  --sbb-form-background-color-readonly: var(--sbb-color-milk);
-  --sbb-form-border-color-readonly: var(--sbb-color-aluminum);
-  --sbb-form-field-width: #{sbb.pxToRem(320)};
-  --sbb-form-input-color: var(--sbb-color-charcoal);
+  --sbb-form-background-color-default: var(--sbb-color-background);
+  --sbb-form-background-color-disabled: var(--sbb-color-milk);
+  --sbb-form-border-color-default: var(--sbb-color-graphite);
```
mhaertwig, line 233: We should not change the color in the standard mode. The original color was charcoal for standard mode. Please change it back.

For lean mode it should be granite, if I see it right. So this has to be overridden in the `where(.sbb-lean)` block.

src/angular/styles/_typography.scss
```diff
@@ -228,11 +228,14 @@
   --sbb-fieldset-legend-nested-font-size: #{sbb.pxToRem(20)};
 
   // Forms
-  --sbb-form-background-color-readonly: var(--sbb-color-milk);
-  --sbb-form-border-color-readonly: var(--sbb-color-aluminum);
-  --sbb-form-field-width: #{sbb.pxToRem(320)};
-  --sbb-form-input-color: var(--sbb-color-charcoal);
```
mhaertwig, line 234: `--sbb-form-input-color` is also used by other components. It needs to be changed there, too.

mhaertwig:
Replaced with #2018 2018

PR #1998, opened 2023-09-13
feat(angular/chips): implement dark mode

antoniafurrer:


src/angular/styles/_typography.scss
```diff
@@ -566,6 +571,12 @@
       --sbb-badge-background-color-disabled: var(--sbb-color-anthracite);
       --sbb-badge-color: var(--sbb-color-black);
 
+      //Chips
```
mhaertwig, line 620:
```suggestion
      // Chips
```

src/angular/chips/chips.scss
```diff
@@ -45,8 +45,8 @@
 
   height: var(--sbb-chips-chip-height);
   border-radius: var(--sbb-chips-chip-height);
-  color: var(--sbb-color-weak-accent);
-  background-color: var(--sbb-color-cloud);
+  color: var(--sbb-chips-color);
```
mhaertwig, line 48: You can leave `color: var(--sbb-color-weak-accent);` and remove `--sbb-chips-color` everywhere.
`--sbb-color-weak-accent` already has the correct colors in standard, lean and dark mode.

mhaertwig (changes requested):

PR #2009, opened 2023-09-19
fix(angular/tabs):  add aria-hidden to inactive tabs

mhaertwig:
Fix accessibility for tabs component. Add `aria-hidden="true"` to inactive tab panels. Fix issue where chromevox would read the names of inactive tab panels when navigating past the active tab panel. Fix this by adding `aria-hidden="true"` to inactive tab panels to exclude them from the a11y tree.


kyubisation (approved):
LGTM 👍

PR #2010, opened 2023-09-19
fix(angular/autocomplete):  requireSelection incorrectly resetting va…

mhaertwig:
…lue when there are no options

The autocomplete has a check not to reset the value if the user didn't interact with the input. The problem was that we only accounted for it when there are options, because while technically an autocomplete is _attached_ when the user focuses, we don't consider it _open_ until it shows some options.


kyubisation (approved):
LGTM 👍

PR #2013, opened 2023-09-19
feat(angular/showcase): implement dark mode

n1thushan:
highlight.js library still needs to be implementet


mhaertwig (approved):
LGTM 👍 
I've updated highlight.js for using the github-dark theme. I like it more because it has a much better contrast.

PR #2019, opened 2023-09-26
fix(journey-maps): hide geolocate button by default

swiss-chris:
Refs: ROKAS-100


mhaertwig:
Isn't it more a `fix` than a `feat`? I am asking because it may affect the version of the next published package.

swiss-chris:
> Isn't it more a `fix` than a `feat`? I am asking because it may affect the version of the next published package.

ja, du hast recht. Reicht es, wenn ich es im PR Titel ändere ?

mhaertwig:
Ja, das reicht aus. Danke

mhaertwig (approved):

PR #2022, opened 2023-09-26
feat(angular/menu): implement dark mode

n1thushan:


src/angular/menu/menu.scss
```diff
@@ -44,6 +44,7 @@ sbb-menu {
   border-radius: var(--sbb-border-radius);
   box-shadow: var(--sbb-box-shadow-below);
   border-top-width: var(--sbb-border-width);
+  border-color: var(--sbb-color-weak-accent);
```
mhaertwig, line 47: Could you try modifying the panel border gloablly in `_typography.scss` instead of here? The select still has the wrong panel color.

src/angular/styles/_typography.scss
```diff
@@ -800,6 +802,8 @@
       --sbb-menu-item-label-active: var(--sbb-color-white);
       --sbb-menu-item-background-color-hover: transparent;
       --sbb-menu-item-background-color-active: var(--sbb-color-charcoal);
+      --sbb-menu-panel-divider-color: var(--sbb-color-graphite);
```
mhaertwig, line 805: Try styling the panel globally. If it looks different across components in Figma, it's probably not intended. We'll later ask Manuel.

mhaertwig (changes requested):

src/angular/styles/_typography.scss
```diff
@@ -2762,6 +2764,7 @@ $panelBorderRadius: 2;
   position: relative;
   top: var(--sbb-box-shadow-horizontal-border-offset-top-below);
   box-shadow: var(--sbb-box-shadow-below), var(--sbb-box-shadow-horizontal-border-bottom);
+  border-color: var(--sbb-color-weak-accent);
```
mhaertwig, line 2767: The border color is already set in line 2757. I suggest to...
* remove the line above and change line 2757 to `border: var(--sbb-border-width) solid var(--sbb-color-weak-accent);`
* remove `--sbb-panel-border-color`

mhaertwig (changes requested):

src/angular/styles/_typography.scss
n1thushan: replacing --sbb-panel-border-color with --sbb-color-weak-accent would change the border color in the dark mode from smoke to milk, is that intended?

src/angular/styles/_typography.scss
```diff
@@ -2780,7 +2780,7 @@ $panelBorderRadius: 2;
       calc(#{sbb.pxToRem(14)} * var(--sbb-scaling-factor));
 
     :where(html.sbb-lean) & {
-      color: var(--sbb-color-granite);
+      color: var(--sbb-menu-title-color);
```
mhaertwig, line 2783: Shouldn't this be placed above the lean block?

src/angular/styles/_typography.scss
mhaertwig: Yes, I think that's what we discussed yesterday. And that's what your current code does.
Maybe I misunderstand something. We can also discuss this in a short call.

mhaertwig (approved):

PR #2043, opened 2023-10-11
refactor: autodetect dark mode

mhaertwig:
This removes the newly introduced `lean_dark` variant. Instead, if the system's preferred color scheme is dark, the dark mode is applied automatically for the lean design variant.

The default behavior can be overridden by adding the `sbb-light` or `sbb-dark` class to the page's `<html>` element.

### DIFF
--- a/src/angular/core/common-behaviors/common.module.ts
+++ b/src/angular/core/common-behaviors/common.module.ts
@@ -3,12 +3,14 @@ import { HighContrastModeDetector } from '@angular/cdk/a11y';
 import { _isTestEnvironment } from '@angular/cdk/platform';
 import { DOCUMENT } from '@angular/common';
 import { Inject, InjectionToken, NgModule, Optional } from '@angular/core';
+import { ɵvariant } from '@sbb-esta/angular/core';
+import { fromEvent, Observable } from 'rxjs';
+import { map, startWith } from 'rxjs/operators';
 
 import { VERSION } from '../version';
 
 import { _global } from './localize/global';
 import { _$localize } from './localize/localize';
-import { ɵvariant } from './variant';
 
 /** @docs-private */
 export function SBB_SANITY_CHECKS_FACTORY(): SanityChecks {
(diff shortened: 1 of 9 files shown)


mhaertwig:
For discussion. I'll add some tests of we agree on that.

src/showcase/app/variant-switch.ts
```diff
@@ -34,27 +36,34 @@ export class VariantSwitch implements CanActivate, OnDestroy {
...
-            document.documentElement.classList.add(`sbb-dark`);
-          }
         }
-        ɵvariant.next(value);
+        ɵvariant.next(value === 'standard' ? 'lean' : 'standard');
         localStorage.setItem(variantLocalstorageKey, value);
+
+        // switch between light and dark mode
+        document.documentElement.classList.remove('sbb-dark');
+        document.documentElement.classList.remove('sbb-light');
```
jeripeierSBB, line 61: You could add multiple args to remove(a, b, ...)

jeripeierSBB, line 61: You could add multiple args to remove(a, b, ...)

jeripeierSBB (approved):
I like the approach!
Little bit disappointed that there is the need of an extra preferred class, but it would be difficult to combine media query with the other class, so it seems the best solution.

jeripeierSBB (approved):
I like the approach!
Little bit disappointed that there is the need of an extra preferred class, but it would be difficult to combine media query with the other class, so it seems the best solution.

src/angular/core/common-behaviors/common.module.ts
```diff
@@ -50,15 +52,17 @@ export class SbbCommonModule {
     @Optional() @Inject(SBB_SANITY_CHECKS) private _sanityChecks: SanityChecks,
     @Inject(DOCUMENT) private _document: Document,
   ) {
+    this._isDarkMode().subscribe((isDark) => {
+      if (isDark) {
+        document.documentElement.classList.add('sbb-preferred-color-scheme-dark');
+      } else {
+        document.documentElement.classList.remove('sbb-preferred-color-scheme-dark');
+      }
```
kyubisation, line 60: Minor nitpick
```suggestion
      document.documentElement.classList.toggle('sbb-preferred-color-scheme-dark', isDark);
```

kyubisation (approved):
LGTM 👍 
I think this is a nice solution. Thank you for your effort! 😃

PR #2063, opened 2023-10-31
chore: update sbb-icon-names

mhaertwig:


mhaertwig:
@kyubisation, @jeripeierSBB: I removed the `generate:icon-list` script from the `integrity` command to avoid failing builds every time the icons are updated. Agree?

jeripeierSBB:
I think it's ok. I'm just wondering why this became a bigger problem than earlier?

jeripeierSBB:
Ah now I understand :) I also once had to fix a commit in journey maps because of this .

jeripeierSBB (approved):
LGTM

PR #2081, opened 2023-11-10
fix(multiple): change dark mode background to midnight and fix some styles

mhaertwig:
This changes the default dark mode background color to midnight. Additionally, some dark mode styles are fixed according to feedback from UX.

### DIFF
--- a/src/showcase/app/variant-switch.ts
+++ b/src/showcase/app/variant-switch.ts
@@ -53,7 +53,7 @@ export class VariantSwitch implements CanActivate, OnDestroy {
         } else {
           document.documentElement.classList.add(`sbb-lean`);
         }
-        ɵvariant.next(value === 'standard' ? 'lean' : 'standard');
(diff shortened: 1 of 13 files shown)


src/showcase/app/variant-switch.ts
```diff
@@ -53,7 +53,7 @@ export class VariantSwitch implements CanActivate, OnDestroy {
         } else {
           document.documentElement.classList.add(`sbb-lean`);
         }
-        ɵvariant.next(value === 'standard' ? 'lean' : 'standard');
+        ɵvariant.next(value === 'standard' ? 'standard' : 'lean');
```
jeripeierSBB, line 56: Was this wrong before?

mhaertwig: Yes, it was a bug in my dark mode pull request.

jeripeierSBB (approved):
LGTM

PR #2101, opened 2023-11-22
fix(showcase): make expandable table example working on all browsers

### LINKED ISSUE #2100: [Bug]: Additional whitespace after each expandable table line on iPhone 14 Pro
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/sbb-design-systems/sbb-angular/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/sbb-design-systems/sbb-angular/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Bug type

Other

### Which version are you using

17.0.0

### Operating system

iOS

### Browser / Browser version

Chrome 92.0.4515.90

### Input mode

Touch

### Additional settings

_No response_

### What happened?

After each expandable table line there is an additional whitespace on iPhone devices, such as iPhone 14 Pro for example (see screenshot). It seams that the space for the expansion is showed under each line from the beginning. I would expect not to see any whitespace, like in the desktop-view or on Android phones, when no line has bean selected.

### Steps to reproduce the issue

1. Open https://angular.app.sbb.ch/angular/components/table/examples on an iPhone 14 pro Browser (for example chrome).
2. Have a look at the expandable table example (the last one)
3. Notice the whitespace under each line
Attached are a screenshot of the issue and one with expected behaviour.
expandableTableWhitespaceIphone14Pro
expandableTableAndroidExpectedBehaviour

### Relevant log output

_No response_

mhaertwig:
Not all browsers support `visibility: collapse` on table rows. Some browsers, e.g. Safari, treat it like `visibility: hidden` which causes a gap below the clickable rows. Therefore, rows are now hidden using `display: none` and shown using `display: table-row` what is supported by all major browsers.

See https://developer.mozilla.org/en-US/docs/Web/CSS/display and https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

Closes #2100

### DIFF
--- a/src/components-examples/angular/table/expandable-table/expandable-table-example.html
+++ b/src/components-examples/angular/table/expandable-table/expandable-table-example.html
@@ -37,7 +37,7 @@
     <tr
       sbb-row
       *sbbRowDef="let row; columns: ['expandedDetail']; when: isExpansionDetailRow"
-      [style.visibility]="row.id == selectedId ? 'visible' : 'collapse'"
+      [style.display]="row.id == selectedId ? 'table-row' : 'none'"
     ></tr>
   </table>
 </sbb-table-wrapper>


kyubisation (approved):
LGTM 👍 
Nice find 😃

jeripeierSBB (approved):
LGTM

PR #2120, opened 2023-12-20
feat(multiple): switch to angular's new built-in control-flow

mhaertwig:
Reworks the library to use the new built-in control flow instead of `ngIf`, `ngSwitch` and `ngFor`. This has the advantage of not requiring directives to be imported and instantiated and increases the overall performance of the library.


mhaertwig:
The first two commits in this PR were created using the automatic migration tool for the angular control-flow. The other PRs are manual changes.

package.json
```diff
@@ -27,7 +27,8 @@
     "test-firefox": "yarn -s test --firefox",
     "tslint": "tslint -c tslint.json --project ./tsconfig.json",
     "lint": "yarn -s tslint && yarn -s bazel:format-lint",
-    "format:prettier": "prettier --write \"**/*.{js,ts,css,scss,json,md,html}\"",
+    "format:prettier": "prettier --write \"**/*.{js,ts,css,scss,json,md}\" && yarn -s format:prettier:html",
+    "format:prettier:html": "prettier --write \"./src/**/*.html\" --parser angular",
```
mhaertwig, line 31: We now use prettier's angular parser for html files to correctly format the new control-flow syntax.

tools/markdown-to-html/transform-markdown.ts
```diff
@@ -28,12 +28,12 @@ if (require.main === module) {
...
+    // We use ```angular code blocks to enable correct syntax formatting with prettier. Since markedjs/marked does
+    // not support custom code block languages, we replace all ```angular code blocks with ```html code blocks.
+    inputFileString = inputFileString.replace(/```angular/g, '```html');
     const outputPath = join(bazelBinPath, inputPath.replace(markdownExtension, '.html'));
-    const htmlOutput = markdownRenderer.finalizeOutput(
-      marked(readFileSync(inputPath, 'utf8')),
-      inputPath,
-    );
-
+    const htmlOutput = markdownRenderer.finalizeOutput(marked(inputFileString), inputPath);
```
mhaertwig, line 36: A bit hacky but prettier breaks the formatting of ```html blocks if they use the new control-flow syntax. That's why I replaced these blocks with ```angular and undo this change here.

jeripeierSBB: but the visual output still looks the same?

PR #2123, opened 2023-12-21
fix(angular/tags): remove focus on mouse blur

### LINKED ISSUE #2119: [Bug]: Is the current active state of the tag component correct?
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/sbb-design-systems/sbb-angular/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/sbb-design-systems/sbb-angular/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Your Question

Is the current active state of the tag component correct?

### Give us a summary about your question

This is a very minor detail - we actually got user feedback about it which is the reason why I'm creating this issue.

When clicking/activating a [tag item](https://angular.app.sbb.ch/angular/components/tag/examples), the border will be gray (which is a focus state i suppose?) but it stays like this until the user clicks anywhere else. If a user clicks 3/5 tags, one will have a different ui state than the two others, effectively leading to 3 different states (default, active and "last active").

Is this done on purpose?

I tried to capture this behavior in a quick screen recording:
https://github.com/sbb-design-systems/sbb-angular/assets/25688064/28bcd7e3-3f3b-4a06-8b3f-217738f8765c

### Provide as much useful information as you can

_No response_

mhaertwig:
The `SbbTag`, which inherits from the `SbbCheckbox`, stayed focussed until the user clicked somewhere else. This PR fixes this and removes the visible focus on blur leaving the element focussed in the background for not breaking keyboard navigation. Only tags focussed using keyboard navigation remain focussed.

Fixes #2119

### DIFF
--- a/src/angular/tag/tag.scss
+++ b/src/angular/tag/tag.scss
@@ -65,7 +65,7 @@
     color: var(--sbb-tag-badge-label-color-disabled);
   }
 
-  .sbb-tag:not(.sbb-tag-disabled):is(:hover, :focus, .cdk-focused) & {
+  .sbb-tag:not(.sbb-tag-disabled):is(:hover, :focus, .cdk-keyboard-focused) & {
     background-color: var(--sbb-tag-background-color-inactive-hover);
     color: var(--sbb-tag-label-color-inactive-hover);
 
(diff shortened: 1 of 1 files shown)


jeripeierSBB (approved):
LGTM.  Thanks

PR #2144, opened 2024-01-23
fix: enforce @angular/cdk versions >= 17.1.0

mhaertwig:
SBB Angular is not working with @angular/cdk versions smaller than 17.1.0. Therefore, we enforce the minimum cdk version 17.1.0.


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #2147, opened 2024-01-23
build: bump year

mhaertwig:


kyubisation (approved):
LGTM 👍 
Should we create an automatic workflow for this? 😄

PR #2155, opened 2024-01-25
fix(stackblitz): update stackblitz assets

mhaertwig:
Increase the default library version to 17.4.1 and CDK to 17.1.0 because older versions are not compatible.


kyubisation (approved):
LGTM 👍

PR #2158, opened 2024-01-29
fix(angular/dialog): sbb-dialog-title should work under OnPush `viewContainerRef`

mhaertwig:
The `sbb-dialog-title` directive updates state in a microtask and should call `ChangeDetectorRef.markForCheck`. Failing to do this will cause the component tree to not be checked if it lives under an `OnPush` component that has not otherwise been marked for check. This commit updates the queue to be a signal so we don't have to think about calling `markForCheck` at the right times.


kyubisation (approved):
LGTM 👍

PR #2161, opened 2024-01-29
fix(multiple): import ANIMATION_MODULE_TYPE from core

mhaertwig:
Moves all imports of `ANIMATION_MODULE_TYPE` from `platform-browser/animations` to `core` to reduce our dependence on the animations module and to avoid potential issues when lazy-loading animations.


kyubisation (approved):
LGTM 👍

PR #2168, opened 2024-01-31
fix(angular/checkbox): move required validation into component

mhaertwig:
Currently, required validation is implemented as a separate directive. This is problematic, because with standalone users would have to add two imports.

These changes move the required validation into the checkbox and deprecate the old module.


kyubisation (approved):
LGTM 👍

PR #2173, opened 2024-02-01
refactor(multiple): extract error tracking behavior into a class

mhaertwig:
Extract error tracking behavior into a class and remove TypeScript mixin usages. The `mixinErrorState` is marked as deprecated.


kyubisation (approved):
LGTM 👍

PR #2174, opened 2024-02-01
refactor(angular/dialog): use enum instead of const enum

mhaertwig:
Const enums prevent enabling the `isolatedModules` check which is needed for fast TS compilation based on transpileModule API.


kyubisation (approved):
LGTM 👍

PR #2176, opened 2024-02-01
fix(angular/autocomplete): clear selected option if it is removed while typing

mhaertwig:
Clears the selected option in the autocomplete if the user types in something else.

This also fixes a regression #2010 where the call to capture the input value was moved into the root of `_attachOverlay` in order to capture it even when there are no options. This is problematic, because `_attachOverlay` is called as the user is typing which basically breaks the `requireSelection` feature.

These changes resolve the issue while preserving the original fix by only capturing the value when the overlay is actually attached.


kyubisation (approved):
LGTM 👍

PR #2193, opened 2024-02-15
fix(angular/tabs): prevent scrolling and disable interaction if disabled

mhaertwig:
Fixes that while we were preventing clicks on disabled links using `pointer-events`, we didn't prevent keyboard events.

Moreover, scrolling when hitting space is prevented.


kyubisation (approved):
LGTM 👍

PR #2194, opened 2024-02-15
fix(angular/schematics): import async animations and remove deprecate…

mhaertwig:
…d functions

* Reworks the `ng add` schematic so it doesn't depend on deprecated functions.
* Imports the async animations instead of the eagerly-loaded ones.


kyubisation (approved):
LGTM 👍

PR #2195, opened 2024-02-16
build: add workflow to regenerate images of latest major versions

kyubisation:
This should prevent security issues from accumulating in the images, as they are continuously regenerated from the newest nginx stable image.


mhaertwig (approved):
LGTM 👍

PR #2210, opened 2024-03-07
fix(journey-maps): fix horizontal level switch animation

andrinsibold:
Fixes the animation of the horizontal level switcher 
Refs: ROKAS-1884


swiss-chris (approved):
LGTM

PR #2212, opened 2024-03-07
refactor(multiple): remove TypeScript mixin usage

mhaertwig:
This removes the usage of `mixinInitialized` and marks it as deprecated


kyubisation (approved):
LGTM 👍

PR #2213, opened 2024-03-11
fix(angular/autocomplete): requireSelection sometimes not clearing value

mhaertwig:
Fixes that if the user has `requireSelection` enabled, selects a value and then deletes a character while the input still has a value, the selection wasn't being cleared as expected.


kyubisation (approved):
LGTM 👍

PR #2214, opened 2024-03-12
feat(journey-maps): enable start level even without leit-pois

besartmemeti:


mhaertwig (approved):

mhaertwig:
@besartmemeti I merged this a bit too fast. Do you agree if I rename this to "fix" instead of "feat"? I am asking because it affects our versioning.

besartmemeti:
@mhaertwig sure, this is perfectly fine for me.

PR #2219, opened 2024-03-26
fix(angular/core): mark fields on HasErrorState as nullable

mhaertwig:
Updates fields on `HasErrorState` to be nullable to align them with the `_ErrorStateTracker`.


kyubisation (approved):
LGTM 👍

PR #2228, opened 2024-04-03
feat(angular/tooltip): allow specifying the preferred tooltip position

### LINKED ISSUE #889: Feature request: Tooltip preferred position
We use the tooltip in the icon bar - there we'd prefer it to appear on the right, so it does not overlap the next icon:

image

A new input to the tooltip - e.g. `preferredPosition: 'top' | 'bottom' | 'left' | 'right'` that is honored if there is enough space available. Otherwise it should fall back to the original/current postitioning.

mhaertwig:
This allows specifying the preferred position of the tooltip. Allowed position values are `above`, `below`, `left`, `right`. Each position has at least one fallback position associated where the tooltip is positioned if there's not enough space.

If the tooltip is positioned `below` or `above`, if is centered by default and moved to the start or end position if there's not enough space to the side.

Closes #889

### DIFF
--- a/src/angular/tooltip/BUILD.bazel
+++ b/src/angular/tooltip/BUILD.bazel
@@ -17,6 +17,7 @@ ng_module(
     ),
     assets = [":tooltip.css"] + glob(["**/*.html"]),
     deps = [
+        "//src:dev_mode_types",
         "//src/angular/core",
         "//src/angular/icon",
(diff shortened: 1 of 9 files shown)


kyubisation (approved):
LGTM 👍 
Thanks! 😃

jeripeierSBB (commented):
LGTM

PR #2235, opened 2024-04-10
chore: adjust deprecation target version

mhaertwig:
Give the consumers a bit more time. These deprecations are only a few weeks old.


kyubisation (approved):
LGTM 👍

PR #2251, opened 2024-05-01
fix(angular/tabs): remove dependency on NgClass

mhaertwig:
We can set classes directly through `[class]` instead of having to import `NgClass`.


kyubisation (approved):
LGTM 👍

PR #2255, opened 2024-05-01
fix(angular/menu): remove dependency on NgClass

mhaertwig:
We can set classes directly through `[class]` instead of having to import `NgClass`. The code had to be adjusted a bit, because the built-in `[class]` binding doesn't do diffing on the object literal.


kyubisation (approved):
LGTM 👍

PR #2268, opened 2024-05-21
fix(angular/tooltip): tooltip should mark for check when visibility changes

mhaertwig:
This commit fixes an error in the tooltip component where it updates a value used in a host binding without calling `markForCheck`. This does not currently throw an error because `checkNoChanges` is broken for `OnPush` components. However, the error was surfaced when testing an "exhaustive" check that will check _all_ views, regardless of their dirty state. The issue also doesn't really produce a _visible_ problem because it occurs when the tooltip is hidden so the fact that the zoom never updates doesn't make a real difference.


kyubisation (approved):
LGTM 👍

PR #2269, opened 2024-05-21
fix(angular/datepicker): resolve repeater warnings in calendar

mhaertwig:
Fixes that the calendar was triggering some newly-introduced warnings from the framework.


kyubisation (approved):
LGTM 👍

PR #2280, opened 2024-05-27
fix: update schematics for v18

mhaertwig:
Updating SBB-Angular to v18 using the recommended `ng update` did not work because the schematic function was not updated. This PR fixes the update schematics.


kyubisation (approved):
LGTM 👍

PR #2286, opened 2024-06-04
docs: remove Oauth documentation part

mhaertwig:
We now recommend using `angular-auth-oidc-client` instead of `angular-oauth2-oidc`. As discussed internally, we're therefore removing this documentation from the SBB Angular showcase. The new documentation can be found in the CLEW confluence.


kyubisation (approved):
LGTM 👍 
While technically breaking, I don't think it is used anywhere, so we can ignore the breaking change.

PR #2307, opened 2024-07-09
fix(angular/dialog): provide dialog service in root

mhaertwig:
This provides the dialog service in root and adds deprecations to unused injection tokens.


kyubisation (approved):
LGTM 👍 
Should we also do this with the lightbox?

mhaertwig:
Good point, thanks. I created another PR: #2308

PR #2308, opened 2024-07-09
fix(angular/lightbox): provide lightbox service in root

mhaertwig:
This provides the lightbox service in root and adds deprecations to unused injection tokens.


kyubisation (approved):
LGTM 👍

PR #2321, opened 2024-07-18
refactor(angular/tabs): remove use of zone.onStable for aligning tabs

mhaertwig:
Also update the lint configuration to the current state.


kyubisation (approved):
LGTM 👍

PR #2326, opened 2024-07-18
fix(angular/sidebar): not closing on escape key press

mhaertwig:
Fixes that the sidebar wasn't triggering change detection when closing via escape key press.


kyubisation (approved):
LGTM 👍

PR #2331, opened 2024-07-23
test: convert remaining tests to zoneless change detection

mhaertwig:
Convert remaining tests to zoneless change detection and use zoneless by default for tests. For testing functionality to work with zoneless change detection, move the test to a `*.zone.spec.ts` file.


kyubisation (approved):
LGTM 👍 
Great work!

PR #2332, opened 2024-07-23
fix(angular/datepicker): avoid losing focus when re-rendering the current view

mhaertwig:
When some calendar inputs change, we need to re-render the entire view (e.g. `minDate` or `dateFilter`). This can cause the active cell to lose focus since it's being re-rendered.

These changes schedule the active cell to be re-focused in the cases where the view would be re-rendered.


kyubisation (approved):
LGTM 👍

PR #2338, opened 2024-08-13
fix(angular/tooltip): remove aria-describedby when disabled

mhaertwig:
Fixes that we were setting an `aria-describedby` even if the tooltip won't show up because it's disabled.


kyubisation (approved):
LGTM 👍

PR #2346, opened 2024-08-26
feat(journey-maps): new input field for journey-routes data

swiss-chris:
Refs: ROKAS-2171


src/journey-maps/angular/journey-maps.ts
```diff
@@ -576,20 +576,17 @@ export class SbbJourneyMaps implements OnInit, AfterViewInit, OnDestroy, OnChang
 
           // remove previous data from map
           this._mapJourneyService.updateJourney(this._map, mapSelectionEventService, undefined);
-          this._mapTransferService.updateTransfer(this._map, undefined);
+          this._mapTransferService.updateTransfer(this._map, undefined); // TODO: asi, still needed for v1?
```
andrinsibold, line 579: Is this needed for v1/ api calls?

src/journey-maps/angular/services/map/map-journey-service.ts
```diff
@@ -33,7 +33,7 @@ export class SbbMapJourneyService {
       const properties = feature.properties!;
       const type = properties.type;
       const pathType = properties.pathType;
-      const legId = properties.legId ?? 'journey'; // default: all belong together
+      const legId = properties.legId ?? 'trip'; // default: all belong together
```
andrinsibold, line 36: With these changes the method & thus the class name could be changed aswell, right?

swiss-chris: In principle, yes. On the other hand, we still have a TransferService and a RouteService class, so maybe better to just leave them all for now for consistency and then rename them when we delete v1 / TransferService ?

src/journey-maps/angular/journey-maps.ts
swiss-chris: I'm confused by the combination of the terms "still" and "v1". Did you perhaps mean "still needed for v2" ?

src/components-examples/journey-maps/angular/journey-maps-routes-zones/journey-maps-routes-zones-example.html
```diff
@@ -6,7 +6,7 @@
     [apiKey]="apiKey"
     language="en"
     #advancedMap
-    [journeyMapsRoutingOption]="journeyMapsRoutingOption"
+    [journeyRoutesOption]="journeyRoutesOption"
```
swiss-chris, line 9: This would be a breaking change, and AFAIK we are trying to implement ROKAS-2171 without any breaking changes.

src/journey-maps/angular/services/map/map-journey-service.ts
swiss-chris: On another note, I'm not very happy with this magic string 'journey' / 'trip'. [wtf](https://www.osnews.com/story/19266/wtfsm/). I realise it's not your code, but maybe we could figure out a cleaner way to do this ? Ideally we should not store a string into a variable which is meant to contain integers. Or at the very least use a constant and make sure the same constant is referred to everywhere the code checks for this value... as is, I can't even review this code change because I don't know what it might break.

src/journey-maps/angular/journey-maps.interfaces.ts
```diff
@@ -74,45 +74,35 @@ export interface SbbBoundingBoxOptions {
 }
 
 /**
- * **WARNING:** The map currently doesn't support more than one of these fields to be set at a time.
+ * **WARNING:** The map doesn't support more than one of these fields to be set at a time.
```
swiss-chris, line 77: @andrinsibold diese Warning ist noch zu unpräzise (war sie vorher auch schon). Müsste so etwas sein wie "The map doesn't support "trip" and "routes" to be set at the same time".

src/journey-maps/angular/services/map/map-journey-service.ts
```diff
@@ -19,21 +19,44 @@ export class SbbMapJourneyService {
     private _mapTransferService: SbbMapTransferService,
   ) {}
 
+  /**
+   * @deprecated
```
andrinsibold, line 23: remove this deprecated for internal methods

src/journey-maps/angular/util/input-validation.ts
```diff
@@ -1,7 +1,7 @@
-import { SbbJourneyMapsRoutingOptions } from '../journey-maps.interfaces';
+import { SbbJourneyMapsRoutingOptions, SbbJourneyRoutesOptions } from '../journey-maps.interfaces';
 
 export const getInvalidRoutingOptionCombination = (
-  routingOptions: SbbJourneyMapsRoutingOptions,
+  routingOptions: SbbJourneyMapsRoutingOptions & SbbJourneyRoutesOptions,
```
mstankala, line 4:
```suggestion
  routingOptions: SbbJourneyMapsRoutingOptions | SbbJourneyRoutesOptions,
```

src/journey-maps/angular/journey-maps.ts
```diff
@@ -596,6 +605,44 @@ export class SbbJourneyMaps implements OnInit, AfterViewInit, OnDestroy, OnChang
       }
     }
 
+    // handle journey, transfer, and routes together, otherwise they can overwrite each other's transfer or route data
+    if (changes.journeyRoutesOption) {
+      const invalidKeyCombination = getInvalidRoutingOptionCombination(
```
andrinsibold, line 610: getInvalidRoutingOptionCombination seperate Methode von Routes / Journey

mstankala (commented):
Add examples files for routes.

src/journey-maps/angular/services/map/map-journey-service.ts
```diff
@@ -69,3 +92,8 @@ export class SbbMapJourneyService {
     }
   }
 }
+
+export enum LegIdType {
```
andrinsibold, line 96: enum not needed.

mstankala (approved):
LGTM, Thx Andrin

mhaertwig (approved):

PR #2349, opened 2024-08-28
fix(angular/form-field): update state if control changes

mhaertwig:
Fixes that the form field wasn't accounting for the case where the form control is swapped out.


kyubisation (approved):
LGTM 👍

PR #2350, opened 2024-08-28
fix(angular/tabs): allow for tablist aria-label and aria-labelledby t…

mhaertwig:
…o be set

According to the [W3C reference implementation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), the inner `tablist` can be labelled using `aria-label` or `aria-labelledby`. These changes add an input to allow them to be set.


kyubisation (approved):
LGTM 👍 
Thanks!

PR #2355, opened 2024-08-29
build: update @angular/bazel to consume new format of APF

mhaertwig:
This also updates Angular to the newest version.


kyubisation (approved):
LGTM 👍

PR #2357, opened 2024-09-03
fix(angular/sidebar): prevent unnecessary events when switching to mobile

mhaertwig:
Fixes an issue where 'opened' or 'closed' events were triggered when switching between mobile and non-mobile mode, regardless of whether the sidebar was already open or closed.


kyubisation (approved):
LGTM 👍

PR #2371, opened 2024-09-11
fix(journey-maps): changed POI V3 style default visibility to visible

mstankala:
Fix: by default all 2nd-level POI (-categories) in the new V3 style should be visible


davidkuehner (approved):
lgtm

PR #2383, opened 2024-09-23
feat(angular/checkbox): add new aria properties to SbbCheckbox

mhaertwig:
Added three new aria properties to SbbCheckbox:
`aria-expanded`: Indicates whether the checkbox controls the visibility of another element. This should be a boolean value (true or false). `aria-controls`: Specifies the ID of the element that the checkbox controls. `aria-owns`: Specifies the ID of the element that the checkbox visually owns.

These attributes will be added to the generated checkbox element if they are specified and won't be present in the HTML if not provided.


kyubisation (approved):
LGTM 👍

PR #2384, opened 2024-09-23
fix(angular/select): remove incompatible aria-autocomplete attribute

mhaertwig:
Fixes a bug where the usage of aria-autocomplete is not a valid attribute that can be used with aria role=listbox. Removes the aria-autocomplete attribute to fix the violation: aria-allowed-attr.


mhaertwig:
Tests will be fixed by #2385

kyubisation (approved):
LGTM 👍

PR #2388, opened 2024-09-24
build: update to Angular 19 (next)

mhaertwig:
BREAKING CHANGE: This updates all Angular packages to the newest version. `standalone: false` was added by the automatic migration. In addition, unused imports are now identified by the builder and were removed manually.


src/angular/core/option/optgroup.ts
```diff
@@ -56,6 +56,7 @@ export const SBB_OPTGROUP = new InjectionToken<SbbOptgroup>('SbbOptgroup');
     '[class.sbb-disabled]': 'disabled',
   },
   providers: [{ provide: SBB_OPTGROUP, useExisting: SbbOptgroup }],
+  standalone: false,
```
kyubisation, line 59: Is there a blocker that prevents us from converting these components to standalone?

kyubisation (approved):
LGTM 👍

src/angular/core/option/optgroup.ts
mhaertwig: No. I wasn't aware they are still not standalone and will convert them in a separate step.

PR #2389, opened 2024-09-24
refactor(multiple): remove unnecessary base classes

mhaertwig:


kyubisation (approved):
LGTM 👍 
Should this be a breaking change and only target the main branch or do you think it's not that important?

mhaertwig:
As discussed, nobody seems to use the base classes which is why we don't mark it as a breaking change.

PR #2393, opened 2024-09-30
fix(angular/tooltip): correct icon styling when placed inside label

mhaertwig:
Fixes the size of the tooltip icon if it's placed inside an `<sbb-label />` that's not wrapped by an `<sbb-form-field />`.

Example before this change: 
image


kyubisation (approved):
LGTM 👍

PR #2405, opened 2024-10-28
build: update to latest Angular next version

mhaertwig:
BREAKING CHANGE: This updates our libraries to Angular 19. Code changes were made by an automatic migration.


kyubisation (approved):
Generally LGTM
There is an overlap with #2406, which probably needs to be resolved

mhaertwig:
I've already resolved them.

PR #2414, opened 2024-11-06
refactor(angular/tabs): remove unnecessary base classes

mhaertwig:


src/angular/schematics/ng-update/migrations/sbb-angular-symbols.json
```diff
@@ -507,8 +507,6 @@
   "SbbTableWrapper": "table",
   "SbbTableWrapperScrollOffset": "table",
   "SbbTextColumn": "table",
-  "_SbbTabLinkBase": "tabs",
```
mhaertwig, line 510: Technically breaking but I did not find any usages.

kyubisation (approved):
LGTM 👍 
Should we target that for version 19, which is going to be released within the next two weeks?

PR #2415, opened 2024-11-06
docs: remove standalone flag from examples

mhaertwig:
As of Angular 19, components are standalone by default.


kyubisation (approved):
LGTM 👍

PR #2418, opened 2024-11-06
fix(angular/processflow): remove IE animation workaround

mhaertwig:
Removes a workaround that's no longer necessary now that we don't support IE.


kyubisation (approved):
LGTM 👍

PR #2422, opened 2024-11-07
refactor(multiple): use ID generator

mhaertwig:
Switches to using the ID generator service to create unique IDs.


kyubisation (approved):
LGTM 👍 
Nice find!

Issue #2408, opened 2024-10-29
[Question]: Compatibility Issues with new strict CSP Rules and SBB Angular Component Library


aram-yesildeniz:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/sbb-design-systems/sbb-angular/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/sbb-design-systems/sbb-angular/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Your Question

Is the SBB Angular Component Library compatible with the new strict Content Security Policy (CSP) rules added in the ESTA Blueprint?

### Give us a summary about your question

We want to integrate the new CSP rules from the ESTA Blueprint: https://code.sbb.ch/projects/KD_ESTA_BLUEPRINTS/repos/esta-cloud-angular/commits/f099676d56ab8ff7456119f13833ec6d6bef410e#docker%2Fnginx-location.conf

Due to restrictions on inline styles, we’ve been unable to resolve the errors and created a minimal example to identify which components are causing the issue.
It appears that some SBB Angular components are involved. For instance, even with an example that only renders an SBB Checkbox, we still encounter this error.
Example repo: https://code.sbb.ch/projects/AMN_NEON/repos/csp-test/browse

Error:
`Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-xxx'". Either the 'unsafe-inline' keyword, a hash ('sha256-xxx'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present.`

Is there a proper solution for this? If not, could we add to the documentation that it’s acceptable to relax the inline style restrictions?

Thanks and regards

aram-yesildeniz:
Hi @mhaertwig, ok great, thanks for the quick check and fast reply, we are looking forward to the new versions :)
Regards Aram

aram-yesildeniz:
Hey guys, sorry to bother you again regarding CSP.
It looks like the @sbb-esta/journey-maps is also conflicting with the strict inline styles CSP.
[...]

PR #2423, opened 2024-11-11
feat(multiple): make SBB Angular CSP compatible

mhaertwig:
This removes all inline styles for making the library CSP compatible.

See #2408


kyubisation (approved):
LGTM 👍 
Thanks, nice work! 😄

PR #2431, opened 2024-11-20
fix(angular/form-field): avoid touching the DOM on each state change

mhaertwig:
Currently we set the `aria-describedby` every time the state of the form control changes. This is excessive, because it only needs to happen if the error state or `userAriaDescribedBy` change.


kyubisation (approved):
LGTM 👍

PR #2432, opened 2024-11-20
fix(angular/menu): handle keyboard events through dispatcher

mhaertwig:
Currently `sbb-menu` handles its keyboard events in the template, however this ignores the overlay's stacking context which can capture some events that it shouldn't.

These changes switch the menu to handling the events through the common dispatcher instead.


kyubisation (approved):
LGTM 👍

PR #2433, opened 2024-11-20
fix(angular): remove final references to ComponentFactoryResolver

mhaertwig:
Removes the final places where we were referring to `ComponentFactoryResolver` in our SBB Angular code.


mhaertwig:
@davidkuehner, @swiss-chris, fyi: There's still a usage of the deprecated `ComponentFactoryResolver` in journey-maps. It will be removed in version 20.

kyubisation (approved):
LGTM 👍

PR #2447, opened 2024-12-02
build: fix preview deployment

mhaertwig:
The call to `github.rest.repos.createDeployment()` fails if any required commit status checks do not have a "success" state (e.g., when "preview-deployment" has a "skipped" state). This change ensures that only relevant contexts are considered, ignoring the others.


mhaertwig:
@kyubisation I think this needs to be merged in `main` to see if it works.

kyubisation (approved):
LGTM 👍 
Thanks! The API is a bit vague about requirements here, so thanks for investigating this. 😃 
And yes, it needs to be on main. 👍

PR #2453, opened 2024-12-09
fix(journey-maps): allow showing default and custom 3d-buildings ...

swiss-chris:
…independently

Refs: ROKAS-2163


davidkuehner (approved):
LGTM

PR #2466, opened 2025-01-07
fix(angular/select): add opt-in input that allows selection of nullable options

mhaertwig:
By default, `sbb-select` treats options with nullable values as "reset options", meaning that they can't be selected, but rather they clear the select's value. This behavior is based on how the native `select` works, however in some cases it's not desirable. These changes add an input that users can use to opt out of the default behavior.


kyubisation (approved):
LGTM 👍

PR #2468, opened 2025-01-08
fix(angular/accordion): prevent focus from entering the panel while it's animating

mhaertwig:
Currently, the expansion panel prevents focus from entering it using `visibility: hidden`, but that only works when it's closed. This means that if the user tabs into it while it's animating, they may scroll the content make the component look broken.

These changes resolve the issue by setting `inert` on the panel content while it's animating. Also cleans up an old workaround for IE.


kyubisation (approved):
LGTM 👍

PR #2472, opened 2025-01-13
fix(angular/menu): remove dependency on animations module

mhaertwig:
This should allow us to avoid some of the pitfalls of the animations module like occasional memory leaks and reduce the bundle size.


src/angular/usermenu/usermenu.scss
```diff
@@ -230,6 +230,31 @@
...
 
+@keyframes _sbb-usermenu-enter {
+  from {
+    width: var(--sbb-menu-trigger-width);
+    max-height: var(--sbb-menu-trigger-height);
+    opacity: 0;
+  }
+  to {
+    width: 288px;
+    max-height: 100%;
```
mhaertwig, line 241: Using `max-height` to animate the user menu's height creates a less fluid animation compared to directly animating its actual height.

In Angular animations, we achieved this smoothly by using `height: '*'`, where the wildcard dynamically applies the element's height before the animation starts.

@jeripeierSBB, do you have any suggestions for improving the CSS transition?

mhaertwig: I got this working by using `grid-template-rows` for the height animation and by adding a `<div class="sbb-menu-panel-content" />` wrapper element around the menu content.

kyubisation (approved):
LGTM 👍 
Nice work! 😃

PR #2475, opened 2025-01-16
fix(angular/sidebar): switch away from animations module

mhaertwig:
Reworks the sidevar to animate using CSS, rather than the animations module. This requires less JavaScript, is simpler to maintain and avoids some memory leaks caused by the animations module.


kyubisation (approved):
LGTM 👍 
Nice work!

PR #2476, opened 2025-01-16
perf(angular/tooltip): use afterNextRender to invoke aria describer

mhaertwig:
This is done to align layout updates with other components using afterNextRender.


kyubisation (approved):
LGTM 👍

PR #2487, opened 2025-02-03
refactor(multiple): use renderer for manual event bindings

mhaertwig:
Switches our manual calls into `addEventListener` to use the renderer instead so that tooling (e.g. the tracing service) gets notified about them.


kyubisation (approved):
LGTM 👍

PR #2497, opened 2025-03-05
build: require at least cdk version 19.1.0

mhaertwig:
Since #2487 SBB Angular uses functionality from `@angular/cdk/platform` that was introduced in `@angular/cdk` 19.1.0. Therefore, we need at least this version.


jeripeierSBB (approved):
You are fast ;-) thank you!

PR #2521, opened 2025-05-20
fix(angular/select): remove value from aria-labelledby

mhaertwig:
This is to avoid values are announced multiple times.


kyubisation (approved):
LGTM 👍

PR #2527, opened 2025-05-22
fix(angular/core): delete deprecated APIs

mhaertwig:
Delete the mixin APIs that have been deprecated for a while and aren't used anywhere.

BREAKING CHANGE:
* `mixinErrorState` has been removed.
* `mixinDisabled` and `CanDisable` have been removed. Use input transforms instead.
* `mixinInitialized` and `HasInitialized` have been removed. Use a `Subject` that emits in `ngOnInit` instead.
* `mixinTabIndex` and `HasTabIndex` have been removed. Use input transforms instead.


kyubisation (approved):
LGTM 👍

PR #2528, opened 2025-05-22
fix(multiple): fix VoiceOver confused by Select/Autocomplete's ARIA semantics

mhaertwig:
For Select and Autocomplete components, fix issues where VoiceOver was confused by the ARIA semantics of the combobox. Fix multiple behaviors:

Fix VoiceOver focus ring stuck on the combobox while navigating options. Fix VoiceOver would sometimes reading option as a TextNode and not communicating the selected state and position in set. Fix VoiceOver "flickering" behavior where VoiceOver would display one announcement then quickly change to another announcement. Fix the same issues for both Select and Autocomplete component.

Implement fix by correcting the combobox element and also individual options.

First, move the aria-owns reference to the overlay from the child of the combobox to the parent modal of the comobobx. Having an aria-owns reference inside the combobox element seemed to confuse VoiceOver.

Second, apply aria-hidden="true" to the ripple element and pseudo checkboxes on sbb-option. These DOM nodes are only used for visual purposes, so it is most appropriate to remove them from the accessibility tree. This seemed to make VoiceOver's behavior more consistent.

### DIFF
--- a/src/angular/select/select.ts
+++ b/src/angular/select/select.ts
@@ -1,4 +1,10 @@
-import { ActiveDescendantKeyManager, LiveAnnouncer, _IdGenerator } from '@angular/cdk/a11y';
+import {
+  ActiveDescendantKeyManager,
+  addAriaReferencedId,
+  LiveAnnouncer,
+  removeAriaReferencedId,
+  _IdGenerator,
+} from '@angular/cdk/a11y';
 import { SelectionModel } from '@angular/cdk/collections';
 import {
   A,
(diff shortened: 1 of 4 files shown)


src/angular/select/select.ts
```diff
@@ -684,6 +693,71 @@ export class SbbSelect
...
 
+  /**
+   * Track which modal we have modified the `aria-owns` attribute of. When the combobox trigger is
+   * inside an aria-modal, we apply aria-owns to the parent modal with the `id` of the options
+   * panel. Track the modal we have changed so we can undo the changes on destroy.
+   */
+  private _trackedModal: Element | null = null;
+
+  /**
+   * If the autocomplete trigger is inside of an `aria-modal` element, connect
```
kyubisation, line 704: Nitpicke; Documentation seems to refer to autocomplete. It should probably be select?

kyubisation (approved):
LGTM 👍

PR #2531, opened 2025-05-22
fix(angular/sidebar): removes -1 tabindex from sidenav in "over" mode

mhaertwig:
Updates the sidebar component so that when the mode is not equal to `side` the tabIndex value is -1. Otherwise, provide a null value.


kyubisation (approved):
LGTM 👍

PR #2543, opened 2025-05-26
fix(angular/radio): assistive technology announcing button as invalid

mhaertwig:
Fixes that some screen readers were reading out "required, invalid data" for radio buttons that were marked as required. The problem was triggered by us using the `required` attribute and the fix is to add `aria-invalid="false"` to it.

Note that an alternate approach is to use `aria-required` instead, but there are some internal checks that require both `aria-required` and `required` to be set which re-introduces the issue.


kyubisation (approved):
LGTM 👍

PR #2544, opened 2025-05-26
fix(angular/radio): updating required value should mark for check

mhaertwig:
This change ensures that updates to the required value of the radio button also mark the component for check, ensuring OnPush compatibility


kyubisation (approved):
LGTM 👍

PR #2554, opened 2025-05-27
ci: fix renovate's post upgrade task

mhaertwig:
There's no equivalent to `--non-interactive` when using `pnpm install`.


kyubisation (approved):
LGTM 👍

PR #2555, opened 2025-06-04
build: update to Angular v20 final

mhaertwig:


mhaertwig:
The failing build will be fixed by the following pull requests.

PR #2556, opened 2025-06-04
fix(angular/processflow): resolve changed after checked errors

mhaertwig:
Resolves several "changed after checked" errors in the processflow by converting internal state to signals.


kyubisation (approved):
LGTM 👍

PR #2564, opened 2025-06-05
chore: add missing constructor export

mhaertwig:
It was unintentionally removed when deleting deprecated apis.


kyubisation (approved):
LGTM 👍

PR #2565, opened 2025-06-05
build: follow symlinks when copying artifacts

kyubisation:
Currently our build script does not follow symlinks when copying artifacts, which breaks in the container build.
This PR adds the configuration to follow symlinks, which should solve this problem.


mhaertwig (approved):
LGTM 👍

PR #2667, opened 2025-10-22
fix(angular/menu): switch internal state to signals

mhaertwig:
Switches the `_isAnimating` state to a signal in order to avoid "changed after checked" errors.


kyubisation (approved):
LGTM 👍

PR #2668, opened 2025-10-22
fix(angular/menu): prevent child menu reopening while parent is anima…

mhaertwig:
…ting away

Fixes that it was possible for the user to re-open a child menu if they quickly hover back to the parent menu while it's animating away.


kyubisation (approved):
LGTM 👍

PR #2703, opened 2025-12-10
fix(angular/menu): remove deprecated animation definitions

mhaertwig:
Delete exported animation definitions that were marked for removal in 21.

BREAKING CHANGE:
* `sbbMenuAnimations` symbol has been removed.


kyubisation (approved):
LGTM 👍

PR #2732, opened 2025-12-16
fix(angular/tooltip): provide scroll strategy in root

mhaertwig:
This is done for making it usable in standalone components without importing the whole module


jeripeierSBB (approved):
LGTM

PR #2734, opened 2025-12-17
chore: remove obsolete takeUntilDestroyed

mhaertwig:
When using `toSignal`, a `DestroyRef` from the current injection context is used making `takeUntilDestroyed` obsolete.


mhaertwig:
Good catch, @jeripeierSBB. Thanks!

PR #2798, opened 2026-05-12
fix(angular/datepicker): allow weekday abbreviations with a period

mhaertwig:
According to https://cldr.unicode.org/, German weekday abbreviations use a period, whereas Swiss weekday abbreviations do not. This caused the datepicker to stop working when `de` was selected as `LOCALE_ID` instead of `de-CH`.

This PR makes the date string parsing more robust.


kyubisation (approved):
LGTM 👍 
Thanks! 😄

PR #2835, opened 2026-07-06
fix(journey-maps): removed unecessary categories

kusimueller:
Removed main categories from the subcategories list, as they are not relevant for the users.


swiss-chris:
LGTM. 
AFAIK, it was an error to include the main categories in this list, as this list is supposed to only contain subcategories. Admittedly, the name `SbbPointsOfInterestCategoryType` as well as the documentation is not clear about this.

swiss-chris (approved):

mhaertwig:
Hi, thanks for the contribution. Could you please prefix your PR title with a type according to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), e.g. `fix` or `docs`? It would also be helpful to make the message more meaningful from an end-user perspective. 

I am also wondering if this PR should be considered a breaking change if removing POI categories can cause TypeScript compilation errors. I didn't check this in detail.

kusimueller:
hope the updates now satisfy your needs.

mhaertwig:
Thank you for adjusting the title! 👍  I've just changed the component to "journey-maps".
@swiss-chris, what do you think regarding the breaking change?

swiss-chris:
> Thank you for adjusting the title! 👍 I've just changed the component to "journey-maps". @swiss-chris, what do you think regarding the breaking change?

I guess theoretically it could be a breaking change if someone hardcoded one of these main category strings into their application and passes it to the journey-maps client. It would be an error on their part to do this however (in the sense that it was never meant to be used like this and has zero effect afaik).

swiss-chris:
I tried adding/selecting "foo" to the list of categories in our https://ki-rokas-showcase-frontend.sbb-cloud.net/routing app and didn't see any errors in the browser console. I'm assuming it's not a breaking change after all.

